PR #346, opened 2026-02-01
fix: bypass input method by mapping key code directly

happyTonakai:
Use direct key code to character mapping instead of UCKeyTranslate. This allows Neru to work correctly regardless of the current input method (Chinese, Japanese, Korean, etc.), as it listens to physical key codes rather than translated characters.

Changes:
- Add keyCodeToCharacter() function for US QWERTY layout mapping
- Remove dependency on keyboard layout and input source
- Support uppercase/lowercase detection via Shift and Caps Lock states

---

<a href="https://app.devin.ai/review/y3owk1n/neru/pull/346">
  <picture>
    Open with Devin
  </picture>
</a>


y3owk1n:
I will go ahead and add some more cases for `36 - Return`, `48 - Tab` and `50 - Tilde`. Also i'll do some formatting shortly.

PR #360, opened 2026-02-07
fix(systray): copy version to use popup instead of osascript notif

y3owk1n:
This PR uses the validation error popup to notify about the version copied instead of osascripts.

https://github.com/user-attachments/assets/5526b3a0-5f63-475e-868e-71a643e55b89

---

<a href="https://app.devin.ai/review/y3owk1n/neru/pull/360" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


y3owk1n:
the more i think about it, i feel like osascript should be just fine, it's nothing important at all, just a confirmation. Let's keep it as it is.

PR #485, opened 2026-03-04
feat(recursive_grid): add auto_exit_actions config option

### LINKED ISSUE #484: Add configurations to support user defined actions that will quit the mode automatically
### Problem description

Right now there are only 2 ways to quit the modes when we are in it.

1. Use the defined quitting shortcut (e.g. "esc")
2. Launch mode with `--action <action>` flag

It would be convenient if we have another way to configure it to quit the mode automatically after action. 

### Proposed solution

Suggestion to add a new key `auto_exit_action = [ ]` by default and let user to configure it by themselves. E.g.

```toml
[recursive_grid]
enabled = true
auto_exit_actions = [ "left_click", "middle_click" ] # user to configure however they like it to behave, default to be empty array
```

This actions should be provided for each mode independently.

To follow up next, maybe after this is implemented, we could have just remove the `--action` flag tho. Seems like this is more useful than `--action` flag.

### Alternatives considered

_No response_

### Screenshots / mockups

_No response_

### Would you like to contribute?

### Additional context

_No response_

tkhorik:
Add configurable auto_exit_actions to recursive_grid mode that allows users to specify which actions should automatically close the overlay after execution.

- Add `auto_exit_actions` field to RecursiveGridConfig
- Add validation for valid action names
- Auto-exit overlay when configured action is executed
- Update documentation

Closes #484


y3owk1n:
@tkhorik ah sorry, i did not notice that you did a PR, i just done my implementation as well, but the main differences is that it supports the same thing for all mode. I think i'll close yours instead, sorry about that ~ 🙏

Refer to #486 instead.

PR #486, opened 2026-03-04
feat: add auto_exit_actions config option for modes

### LINKED ISSUE #484: Add configurations to support user defined actions that will quit the mode automatically

y3owk1n:
Add a new configuration option 'auto_exit_actions' that allows users to
configure which direct action keys should cause the mode to exit
automatically after execution. This provides an alternative way to quit
modes beyond using the escape key or --action flag.

The option is available for hints, grid, and recursive_grid modes.

Fixes #484

---

To follow up on this, I think we should also remove the --action flag implementation.

---

<a href="https://app.devin.ai/review/y3owk1n/neru/pull/486" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/internal/app/services/action_service.go
+++ b/internal/app/services/action_service.go
@@ -189,7 +189,8 @@ func (s *ActionService) HandleActionKey(
 	key string,
 	mode string,
 ) (bool, error) {
-	act, logMsg, ok := s.getActionMapping(key)
+	act, logMsg, _, ok := s.getActionMapping(key)
+
 	if !ok {
 		s.logger.Debug("Unknown action key",
 			zap.String("mode", mode),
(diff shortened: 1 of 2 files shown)


y3owk1n:
> I'm waiting for this!

Fixing a little bit more docs related thingy and it should be merge shortly.

PR #717, opened 2026-04-20
feat(recursive-grid): add training mode

tkhorik:
## Summary
This PR adds a lightweight recursive-grid training mode for memorizing the top-level recursive-grid key layout without changing Neru's existing architecture or normal recursive-grid behavior.

The feature is intentionally small in scope:
- it reuses the existing recursive-grid mode and overlay
- it is available from both the CLI and the systray
- it reads all recursive-grid layout data from user config
- training-specific integers fall back to defaults when not present
- it does not introduce a new subsystem, persistence layer, timer engine, or broad refactor

## User-facing functionality
Users can now start training in two ways:
- `neru recursive_grid --training`
- Systray -> `Activate Modes` -> `Recursive Grid Training`

Training behavior:
- uses the configured top-level `[recursive_grid]` layout (`grid_cols`, `grid_rows`, `keys`)
- chooses one target cell at a time and highlights it in the overlay
- expects the user to press the configured key for that highlighted cell
- increments the target cell's training progress on a correct answer
- applies a configurable penalty on a wrong answer
- hides labels for cells that have reached the configured learning threshold
- finishes automatically when all cells reach the threshold
- supports restart via `Space` or `Backspace`
- exits via `Escape`
- shows a completion notification when training is finished

## Configuration
This PR adds a new nested config block:

```toml
[recursive_grid.training]
enabled = true
hits_to_hide = 3
penalty_on_error = 1
```

Behavior details:
- `enabled`: enables the training entry points
- `hits_to_hide`: number of correct answers needed before a cell label is hidden
- `penalty_on_error`: number of hits removed from the active target after a wrong answer

Default values are wired into `DefaultConfig()` so missing integer values fall back correctly.
Validation rejects:
- `hits_to_hide < 1`
- `penalty_on_error < 0`

### 1. Added training state for recursive-grid
- introduced a small `TrainingSession` in `internal/app/components/recursivegrid/training.go`
- stores per-cell hit counts, current target, last target, and the configured rules
- avoids immediate repeats where possible
- exposes overlay-ready keys so learned cells can be rendered as hidden labels

### 2. Integrated training into the existing recursive-grid mode
- extended recursive-grid activation to accept a `training` flag
- starts a training session only when the flag is enabled in config
- keeps normal recursive-grid behavior unchanged when training is not active
- routes recursive-grid key presses through the training session when active
- exits the mode automatically on training completion
- resets training cleanly on restart/backspace and during mode cleanup

### 3. Added CLI and IPC support
- added `--training` to the `recursive_grid` CLI command
- rejects invalid combinations such as `--training` with `--action`
- passes the option through IPC so the daemon can activate training mode
- returns a training-specific activation message from the IPC handler

### 4. Added systray support
- added a dedicated `Recursive Grid Training` systray menu item
- disables the menu entry when recursive-grid training is disabled in config
- added `App` interface methods to query training availability and trigger activation

### 5. Updated overlay rendering for training
- extended recursive-grid draw calls to accept a `matchedIndex`
- marks the active training cell as matched so the current target is visually emphasized
- hides next-depth preview during training because the exercise is for the top-level layout only
- uses stronger matched colors for the active training cell
- fixed shared overlay state leakage by resetting `hideUnmatched` when entering recursive-grid/training and when cleaning up grid mode

### 6. Added tests
- config validation tests for the new training settings
- unit tests for `TrainingSession` behavior
- systray test updates for the new training action
- recursive-grid mode tests covering the overlay-state reset and existing recursive-grid behavior affected by the change

## Scope and non-goals
This PR intentionally does not add:
- timer-based training windows
- marathon / zen / reinforcement modes
- persistence of training progress
- architectural refactors
- changes to the normal recursive-grid navigation flow beyond the optional training path

## Verification
I verified the feature with focused Go test runs:
- `go test ./internal/app/components/recursivegrid`
- `go test ./internal/app/modes -run 'TestCleanupGridModeResetsHideUnmatched|TestResetCurrentMode_RecursiveGridPreservesHoldMode|TestHandleRecursiveGridKey_CompleteSelectionDoesNotMoveWhenCursorFollowSelectionDisabled'`
- `go test ./internal/app -run '^$'`
- `go test ./internal/cli -run '^$'`

### DIFF
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -150,6 +150,11 @@ max_depth = 10                              # Maximum recursion levels (1–20)
 enabled = false                            # Opt in to native depth transition animation on supported platforms
 duration_ms = 180                          # Recursive-grid animation duration in milliseconds
 
+[recursive_grid.training]
+enabled = true                             # Allow `neru recursive_grid --training` and the systray training entry
+hits_to_hide = 3                           # Hide a cell label after this many correct answers
+penalty_on_error = 1                       # Hits removed from the current target after a wrong key
+
 [recursive_grid.hotkeys]
 "Escape" = "idle"
 "`" = "toggle-cursor-follow-selection"
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -4,11 +4,21 @@ import (
 	"context"
 
 	"go.uber.org/zap"
+
+	"github.com/y3owk1n/neru/internal/app/modes"
 )
 
 // ActivateMode activates the specified mode.
 func (a *App) ActivateMode(mode Mode) {
 	a.modes.ActivateMode(mode)
+}
+
+// ActivateRecursiveGridTraining starts recursive-grid training mode using the
+// configured top-level recursive-grid layout.
+func (a *App) ActivateRecursiveGridTraining() {
+	a.modes.ActivateModeWithOptions(ModeRecursiveGrid, modes.ModeActivationOptions{
+		Training: true,
+	})
 }
 
 // IsFocusedAppExcluded checks if the focused app is excluded.
--- a/internal/app/app_getters.go
+++ b/internal/app/app_getters.go
@@ -58,6 +58,13 @@ func (a *App) RecursiveGridEnabled() bool {
 	return cfg != nil && cfg.RecursiveGrid.Enabled
 }
 
+// RecursiveGridTrainingEnabled returns true if recursive-grid training is enabled.
+func (a *App) RecursiveGridTrainingEnabled() bool {
+	cfg := a.configSnapshot()
+
+	return cfg != nil && cfg.RecursiveGrid.Enabled && cfg.RecursiveGrid.Training.Enabled
+}
+
 // Config returns the application configuration.
 func (a *App) Config() *config.Config {
 	return a.configSnapshot()
--- a/internal/app/app_types.go
+++ b/internal/app/app_types.go
@@ -23,10 +23,11 @@ type Mode = domain.Mode
 
 // Mode constants from domain package.
 const (
-	ModeIdle   = domain.ModeIdle
-	ModeHints  = domain.ModeHints
-	ModeGrid   = domain.ModeGrid
-	ModeScroll = domain.ModeScroll
+	ModeIdle          = domain.ModeIdle
+	ModeHints         = domain.ModeHints
+	ModeGrid          = domain.ModeGrid
+	ModeRecursiveGrid = domain.ModeRecursiveGrid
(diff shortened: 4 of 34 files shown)


y3owk1n:
I just tested this, is this really useful as training tho? It adds near 1000 lines of code and creates more maintenance in the future for this, especially the code is more complicated with cross platform support now. Just thinking out loud here:

If we are going to do training, i am expecting it to be something more like 'providing virtual clickable element', and user need to click the 'virtual element' with their configured grid. Also I think these kind of training does not need to be provided out of the box under neru binary tho. If it's just another overlay to make sure you click the points right, it can be in another project for user to install to do the click training. 

However, this is already achievable with websites like https://humanbenchmark.com/tests/aim. So I am not really sure if this is something we really needs. The core of training is just that we have a place and can keep practicing our configured keys to press things properly. Let's discuss more.

tkhorik:
I see your point, and I mostly agree.
For now, this is just an MVP experiment. The goal is to run a few iterations to refine both the idea and the implementation, not to introduce a fully polished feature yet.
My initial thinking was to evolve it in stages:
Learn and memorize key positions
Gradually remove visual hints (hide letters)
Introduce random targets on the screen that the user needs to “hit” using their grid
So at this stage, it’s more about validating the training concept than committing to a long-term solution.
Also, this PR doesn’t really require approval right now. It can sit without being merged for some time. I don’t currently have enough free time to iterate on it consistently anyway.
And yes, longer term, it probably makes sense to extract this into a separate project, something like “neru the game,” rather than keeping it inside the core binary.
For now, please don’t merge it.

y3owk1n:
> I see your point, and I mostly agree. For now, this is just an MVP experiment. The goal is to run a few iterations to refine both the idea and the implementation, not to introduce a fully polished feature yet. My initial thinking was to evolve it in stages: Learn and memorize key positions Gradually remove visual hints (hide letters) Introduce random targets on the screen that the user needs to “hit” using their grid So at this stage, it’s more about validating the training concept than committing to a long-term solution. Also, this PR doesn’t really require approval right now. It can sit without being merged for some time. I don’t currently have enough free time to iterate on it consistently anyway. And yes, longer term, it probably makes sense to extract this into a separate project, something like “neru the game,” rather than keeping it inside the core binary. For now, please don’t merge it.

Cool, no problem ~ Lets keep it first and see what's next.

y3owk1n:
I don't think this will ever be a thing for this project, shall we close it?

tkhorik:
Yes, I agree.
I don't think this fits the scope or direction of the project anymore, so closing it makes sense.

Issue #698, opened 2026-04-15
Sticky modifier support on supported linux backends


y3owk1n:
On macOS, we supports sticky modifier, and it should also be supported on linux backends.

PR #718, opened 2026-04-21
fix: make Linux indicators follow cursor

### LINKED ISSUE #699: Mode / Sticky Indicator should follow mouse
On macOS, it will follow the mouse when you manually move it, and linux should be implemented the same.

tkhorik:
## Summary
- render Linux mode and sticky modifier indicators on dedicated overlay layers
- clear and redraw indicator layers independently so cursor polling can move badges without trails
- keep wlroots indicator layers passive so they do not steal keyboard focus

Fixes #699.
Related to #698: this covers the Linux sticky indicator rendering piece, but not full sticky modifier input support.

## Tests
- go test ./internal/app/modes ./internal/ui/overlay ./internal/core/infra/overlay
- GOOS=linux CGO_ENABLED=0 go test -exec=true ./...

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,120 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+**Neru** (練る — "to refine through practice") is a keyboard-driven navigation tool for macOS, Linux, and Windows. It lets users navigate the screen and interact with UI elements (click, scroll, drag) using only keyboard input.
+
+- **Language:** Go 1.26.1+ with Objective-C for macOS native APIs
+- **Build tool:** [Just](https://github.com/casey/just) (`justfile` at root)
+- **macOS:** Fully stable; **Linux:** X11 and Wayland (wlroots); **Windows:** Stub/foundations only
+
+## Common Commands
+
+```bash
+# Build
+just build              # Development build → ./bin/neru
+just release            # Release build (optimized, stripped)
+
+# Test
+just test               # All unit tests
+just test-unit          # Unit tests only
+just test-integration   # Integration tests (platform-specific)
+just test-foundation    # Cross-platform foundation tests
+just test-race          # Tests with race detector
+
+# Lint & Format
+just lint               # golangci-lint + clang-format check
+just fmt                # Format Go (goimports, gofumpt, golines) + Objective-C
+just fmt-check          # Check formatting without modifying
+just vet                # go vet
+
+# Pre-commit checklist
+just fmt && just lint && just test && just build
+
+# Other
+just clean              # Remove build artifacts
+just deps               # Download and tidy dependencies
+just bundle             # Create macOS .app bundle
+just generate-all-protocols  # Fetch Wayland protocol XMLs and generate bindings
+```
+
+### Running a single test
+
+```bash
+go test ./internal/core/domain/grid/... -run TestGridSubdivide
+```
+
+## Architecture
+
+Neru uses **Hexagonal (Ports & Adapters) architecture** with strict layer separation:
+
+```
+cmd/neru/             → Entry point
+internal/
+  app/                → Application layer: orchestration, use cases
+    modes/            → Navigation mode implementations (Hints, Grid, Scroll, RecursiveGrid)
+    services/         → Business logic services (HintService, GridService, etc.)
+    components/       → UI components per mode
+  core/
+    domain/           → Pure business logic, no external deps
+      action/         → Action definitions
+      element/        → UI element representation
+      grid/           → Grid subdivision algorithms
+      hint/           → Hint generation
+      state/          → App state, cursor state, modifier state
+    ports/            → Interface contracts (AccessibilityPort, OverlayPort, etc.)
+    infra/            → Concrete platform implementations
+      platform/
+        darwin/       → macOS (Accessibility API, EventTap, Hotkeys, Overlay)
+        linux/        → X11 / Wayland adapters
+        windows/      → Stubs
+      ipc/            → Unix socket IPC
+      eventtap/       → Global keyboard capture
+  ui/overlay/         → Overlay renderer and coordinate system
+  config/             → TOML config loading and validation
+  cli/                → Cobra-based CLI commands
+```
+
+### Navigation Modes
+
+All four modes implement `Mode` interface (`HandleKey`, `HandleActionKey`, `Activate`, `Exit`, `ToggleActionMode`, `ModeType`). The central orchestrator is `internal/app/modes/handler.go`.
+
+1. **Hints** — Overlay labels on clickable elements via macOS Accessibility APIs
+2. **Grid** — Universal coordinate grid, subdivide by typing letters
+3. **Scroll** — Vim-style scrolling (`j`/`k`, `gg`/`G`, `d`/`u`)
+4. **Recursive Grid** — Iteratively subdivides a grid cell (recommended default)
+
+### IPC
+
+Daemon and CLI communicate via Unix sockets. `neru launch` starts the daemon; other commands send messages over IPC with a 5-second timeout.
+
+## Cross-Platform Rules
+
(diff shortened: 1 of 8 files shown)


y3owk1n:
For this implementation, i have tested it out, but I can't make it work as the PR described. When i launch a mode with badge overlay (e.g. scroll mode), the badge shows, but it doesn't follow the cursor when i manually move my cursor nor when the cursor position changes during grid navigation.

Also i see that we are setting keyboard interactivity to NONE for badge overlay, based on my shallow understanding on wayland, i dont think it will work fine in this case, at cases where on scroll mode, we should need "EXCLUSIVE" interactivity to at least capture the keyboard events so that we can perform scroling keys.

---

I have to be honest that i dont have much knowledge in this, i can only verify whether it works or not.

y3owk1n:
I think for this, let's wait for #719 to merge first, only we come back to this to continue iterate. I think #719 is a pre-requisite for this to work properly, and probably you will need to fix lots of git conflicts if both this and #719 going along together. My suggestion is to stop iterating this PR first until #719 is merged.

y3owk1n:
#719 is done, good news, we just enable hyprland support as well because of that. Next i think if you want, we can rebase this one and continue iterate.

y9san9:
To be honest this raises kind of an interesting point. This PR looks purely AI-generated to me, so sorry if it's not. But the point I am talking about is that:

- Right now @tkhorik proposed 300 LoC changes to the codebase via LLMs
- @y3owk1n tested it and found that it doesn't work, told his feedback to @tkhorik 
- @tkhorik will pass this back to LLM to fix, probably

It feels that @tkhorik impact is very low if not negative. Basically @y3owk1n prompts LLM via @tkhorik, but he could do that without @tkhorik if he wanted to. I think today primary contribution is not the code itself, but the ability to take responsibility for that code.

That ability includes but not limited to thorough testing, thinking about how this would evolve in the future and how it will affect current system (also related to #717). Which is not something we can see in that PR.

I am sorry in advance if that sounded offensive, I can assure you that it's not what I mean in that message. It's more of abstract thinking out loud and some reasoning on how it can be improved.

y3owk1n:
@y9san9 good point, honestly I can see that clearly this is a AI prompt PR, and I am not against any AI written code, especially on something that I don't know much about. Will this PR get merged is another story, but at least if it works, I'll look into the detail of the implementation later and try to fix it my way. 

A good example is the #719, I ended up reverting some stuffs that implemented by AI and make it work with my way. 

Of course if possible, I would like the PR to be fully reviewed by submitter before the submit it, but if it's not the case, I am happy to come in to fix them with the ability I can to make sure we can land this properly. 

For now, the project is small, and there's not a lot of contributor. I think that's something I can still manage, but of course it won't be healthy in the long run where I need to come in and re-fix most of the implementation and merge it. 

Thanks for pointing those things out. I hope that in future we will have more well prepared PR so that I can just test and comment on it instead of hands on fixing them or at least making thing works.

tldr, I don't mind AI code, but it should be at least a working code to start with. 

Also, no worries, it doesn't sounds offensive to me at all, this is the truth that you are speaking of. I think the way forward that I would like is to have a linux based core maintainer that will handle most of the linux related code changes (cos I am a noob in it and takes a lot of research time for me), same goes to windows in the future.

tkhorik:
Thanks for the feedback @y9san9, that’s fair.

You’re right that this PR was initially bootstrapped with LLM assistance, and I agree that it doesn’t currently meet the bar for validation or ownership.

For context, I am not able to work on these tasks continuously, and I put this together during a few spare time slots. Since then, I have not had time to return to it, finish it properly, or re-validate the behavior.
I also did not consider the PR to be finished yet, so I was not expecting it to be reviewed for immediate merge. My assumption was that it would give me time to iterate further or potentially allow others to build on top of it.

y3owk1n:
Let's have the line straight, where when it's not a PR that is ready to be reviewed, please make sure they are all draft PR, so that I would skip them instead. Also to not confuse anybody else that looking into the repo.

y9san9:
If that was not ready for review, it changes things. But yeah, kind of confusing to see a Pull **Request** that is not actually a request. Drafts are fine, but usually you would use them in corporate circumstances to show on what you are working right now or if you want to invite someone to collaborate on something big.

In current case it is probably best to stay local until tested or in a fork. Draft PRs are more convenient than just PRs, but I see no reason to spam with Draft PRs as well.

y3owk1n:
> If that was not ready for review, it changes things. But yeah, kind of confusing to see a Pull **Request** that is not actually a request. Drafts are fine, but usually you would use them in corporate circumstances to show on what you are working right now or if you want to invite someone to collaborate on something big.
> 
> In current case it is probably best to stay local until tested or in a fork. Draft PRs are more convenient than just PRs, but I see no reason to spam with Draft PRs as well.

Fair point, and well said, agreeing with my hands and legs.

tkhorik:
I ran Cursor over everything mentioned in this bug, and it rechecked and fixed it. I also reviewed it myself and didn’t find anything else.
It looks ready now and is waiting for your @y3owk1n magic testing 🙂

### 1. Mode indicator badge ignored cursor position (critical)
`modeIndicatorAnchorLocked` was running through the same grid/recursive-grid selection-point logic as the sticky indicator, causing the mode badge to pin itself to the grid selection point instead of following the cursor. Fixed by simplifying the function to always return `cursorPoint`.

### 2. Wayland passive overlay `Hide()` was destructive (medium)
Indicator layers were calling `Hide()` on every tick when no modifiers were active. On Wayland, `neru_wayland_overlay_hide` destroys the layer surface, wl_surface, buffer, and Cairo context — so the next `DrawBadge` call had to call `setup_buffers` + a blocking `wl_display_roundtrip`, causing visible stutter (~16 ms) on every modifier release. Fixed by adding `neru_wayland_overlay_detach` (attaches a NULL buffer and commits without destroying surfaces) and a `HideContent()` Go method that uses it. Indicator hide paths now call `HideContent()` instead of `Clear()+Hide()` for Wayland layers.

### 3. Passive Wayland overlays could incorrectly claim keyboard focus
The `keyboard_interactivity_set` field in `NeruWaylandOverlay` was always initialized to `EXCLUSIVE` regardless of whether the overlay was created as passive (`keyboard_interactive=0`). Fixed by initializing it based on `keyboard_interactive`, so passive indicator layers always use `NONE` interactivity and never steal keyboard focus from the main overlay.

y3owk1n:
Thanks for the effort, is this tested on actual linux? I don't seems to be able to see any differences yet, both builds on the latest commit of this PR.

macOS (working):

https://github.com/user-attachments/assets/c7c769b8-8acc-4956-a8ba-9ae9d10e679a

linux (niri + wayland) (not working properly):

https://github.com/user-attachments/assets/c6095550-cafe-4ef7-8183-cf88d3450ea2

--- 

If possible, please try to test it out on a linux VM first, you can use UTM or any other software to get a working linux to test with. My current setup is using UTM with fedora 44.

tkhorik:
I tried testing it on Ubuntu 24.04.3 LTS, but I ran into some issues with launching the app. I’ll move the PR back to draft for now, continue investigating, and post an update once I’ve verified everything.

tkhorik:
Tested locally on an Ubuntu VM, but results are unreliable.
The VM environment is likely misconfigured, and I didn't have time to fully set up the testing environment.

@y3owk1n could you please verify this on your environment?
I believe the issue should be fixed and working correctly now.

y3owk1n:
I just tried, I don't feel any different compared to the previous iteration tho, it behaves the same when I am testing. Also please do not commit claude.md.

I am not sure, but maybe our virtual pointer is fundamentally flawed or we have not gotten the right way to do mouse syncing properly yet. We should do more research on other projects and study what is the right way instead of keep iterating on top of something might be broken or flawed.

I do not know anything much about apps on linux that has badge following mouse, but i do notice that fedora workstation installation process does have this badge following mouse kind of animation, maybe it's something we can drill in?

Let's not to rush to build on top of something that might be flawed. Step back and take some time to understand the details might be a wise step. Open source is a looooooong way process.

y3owk1n:
I have added support for this at #920, lets close this now ~

PR #719, opened 2026-04-21
feat: support sticky modifiers on supported linux backends

### LINKED ISSUE #698: Sticky modifier support on supported linux backends

tkhorik:
## Description

This PR adds support for sticky modifier just like macOS variant, however in order to support modified clicks, we are required to switch our implementation with utilizing `evdev`, but this also mean that user has to add them self into the input group.

Note that we also have a fallback when input detection failed, it will just use the normal overlay input implementation, where normal click will work fine, but not modified clicks.

## Related Issues

Fixes #698

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

## General Checklist

## Screenshots / Recordings

N/A

## Additional Context

N/A

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -626,6 +626,19 @@ enabled = true
 tap_max_duration = 300
 ```
 
+On Linux, the sticky modifier indicator renders the symbols `❖⇧⌥⌃`. If they
+appear as `[][][][]`, the selected font does not include those glyphs. Set
+`sticky_modifiers.ui.font_family` to a font on your system that can display
+those characters.
+
+```toml
+[sticky_modifiers.ui]
+font_family = "Your installed symbol-capable font"
+```
+
+You can quickly test a candidate font by pasting `❖⇧⌥⌃` into a text editor and
+checking whether it renders correctly.
+
 ---
 
 ## Theme Palette
@@ -744,6 +757,10 @@ defaults read com.apple.HIToolbox AppleEnabledInputSources
 ## Font Configuration
 
 Use `font_family` in any UI section (`hints.ui`, `grid.ui`, `recursive_grid.ui`, `mode_indicator.ui`, `sticky_modifiers.ui`). An empty string uses the system default.
+
+If the sticky modifier indicator shows `[][][][]` on Linux, your current font is
+missing the modifier glyphs `❖⇧⌥⌃`. Set `sticky_modifiers.ui.font_family` to a
+font installed on your system that renders those symbols correctly.
 
 ---
 
--- a/docs/LINUX_SETUP.md
+++ b/docs/LINUX_SETUP.md
@@ -343,3 +343,20 @@ If the setup is correct, Neru should log:
 ```text
 Using Wayland evdev keyboard capture
 ```
+
+### Sticky modifier indicator shows `[][][][]`
+
+The sticky modifier overlay uses Unicode modifier symbols on Linux:
+`❖⇧⌥⌃`. If you see square boxes instead, the configured font does not include
+those glyphs.
+
+Set a font explicitly in your config:
+
+```toml
+[sticky_modifiers.ui]
+font_family = "Your installed symbol-capable font"
+```
+
+An empty `font_family` uses the system default, which may not have the required
+symbol coverage. A quick way to verify a candidate font is to paste `❖⇧⌥⌃` into
+a text editor and confirm the symbols render there first.
--- a/docs/LINUX_SETUP.md
+++ b/docs/LINUX_SETUP.md
@@ -24,6 +24,44 @@ Neru provides native Linux support through two display server backends:
 
 ---
 
+## Wayland Keyboard Capture Permissions
+
+On wlroots-based Wayland compositors, Neru uses direct `evdev` keyboard capture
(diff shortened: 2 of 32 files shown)


y3owk1n:
I am trying to test this out on my wayland machine, there are 2 things i notice:

1. Applying sticky modifier works fine, but when performing actions like `click`, the modifier doesn't seems attached on the click like how our macOS does (maybe we need virtual-keyboard from wayland too?)
2. When applied several sticky modifier, when we are left with last modifier, pressing it doesn't remove it from the badge

I think if we can fix both of this, this is a good merge.

Note that i am only testing on wayland with niri, not sure about X11.

---

I have to be honest that i dont have much knowledge in this, i can only verify whether it works or not.

tkhorik:
Addressed the Wayland feedback in `4f006f6`:

- Added wlroots `zwp_virtual_keyboard_manager_v1` support and wired sticky modifier `PostModifierEvent` through it on Wayland, so sticky modifiers are physically held and should apply to virtual-pointer clicks.
- Tracked and cleared the previous Linux sticky badge rectangle when symbols change or become empty, so toggling off the last modifier should clear the badge.
- Applied the Greptile cleanup notes: removed the redundant Wayland state guard and avoided the subtle in-place slice compaction.
- Fixed the existing no-cgo overlay stub field issue; `GOOS=linux CGO_ENABLED=0 go test -exec=true ./...` now passes.

Tests run:
- `go test ./internal/app/modes ./internal/core/infra/eventtap ./internal/core/infra/accessibility ./internal/ui/overlay ./internal/core/infra/overlay`
- `GOOS=linux CGO_ENABLED=0 go test -exec=true ./...`
- `git diff --check`

y3owk1n:
thanks for the effort, this is not an easy thing to achieve, especially I have not much knowledge in there. After this, i notice some other issue:

1. when a sticky modifier is applied, it seems like it also applies to keyboard presses. Unlike our macOS version, say `ctrl` is added to sticky modifier, and say we have `gcrhtn` as our recursive grid keys, pressing those key results in `ctrl+g`, `ctrl+c`, and for my case, i'v mapped `ctrl+c` as exit key, so when i add `ctrl` as sticky, and press `c` key, it exits.
2. on linux, ctrl+left_click should result in opening a new tab on browser, i am not entirely sure what's wrong, but it didn't work

Showing some screen record here (sorry for the bad recording, i am still very bad at linux keybindings...lol):

1. first attempt is to show using manual `ctrl` + manual `mouse click` will open a new tab
2. second attempt is that i added `ctrl` as sticky, but pressed `c` key which exited my mode :(
3. third attempt is that i made sure i am on the location to press, activated `ctrl`, and then `left_click`, but it didn't open a new tab, not sure why.

https://github.com/user-attachments/assets/babdb956-292a-43aa-955c-b94631976bd5

tkhorik:
Updated this in `a01e14e`.

What changed:
- Wayland no longer presses the virtual keyboard modifier when sticky state is toggled on. That avoids leaking sticky Ctrl/Alt/etc into Neru navigation keys like recursive-grid `c`.
- Key dispatch now uses only the sticky-stripped key when a sticky modifier is active, so sticky Ctrl + `c` cannot match a configured `Ctrl+C` hotkey.
- wlroots pointer actions now press virtual-keyboard modifiers immediately around virtual-pointer clicks, then release them. Split mouse-down/mouse-up remembers the modifiers held on mouse-down and releases those on mouse-up/cleanup.
- Added a regression test for sticky Ctrl + `c` not triggering the configured `Ctrl+C` hotkey.

Tests run:
- `go test ./internal/app/modes ./internal/core/infra/eventtap ./internal/core/infra/accessibility ./internal/ui/overlay ./internal/core/infra/overlay`
- `GOOS=linux CGO_ENABLED=0 go test -exec=true ./...`
- `git diff --check`

y3owk1n:
Thanks ~ looks like we are one step nearer to merge! A couple more issue i found on testing:

1. set `ctrl` to sticky key, go to a point, do `left_click`, the first click is with `ctrl`, and then we do another `left_click`, now the `ctrl` was removed from the sticky after the first click is done.
2. set `ctrl` to sticky key, go to a browser and press on a link, it should open a new tab (ctrl+click), but it just open on the current tab. This tells me that something might be wrong on the clicking layer that doesn't consider the modifiers properly or maybe our mouse action implementation has some problem in it. I honestly am not sure why it didn't work for the sticky key on clicking tho... hmmm

The rest seems working fine for me other than the above 2 issue.

y3owk1n:
after some research, i think the reason is because the focus is on the overlay with EXCLUSIVE interactivity, the click passes through, but our modifier doesnt pass through, even tho the console says we passed the modifiers, still figuring what's next.

y3owk1n:
I am able to make it work for now, maybe later i will commit it here and we can see what's next. The solution is that we have to actually use evdev instead of relying on overlay inputs...

also we might possibly accidentally support hyprland too, i'll confirm later.

y3owk1n:
i think we are good to merge after CI passes

y3owk1n:
Thanks @tkhorik for starting this!

Issue #696, opened 2026-04-15
Hyprland not working


y3owk1n:
We have added support on wayland with wlroots, can confirm that it works on sway and niri, but not hyprland. I tried various of fix, but could not make it happen at all. There's a sort of working workaround that can be found at branch `fix/hyprland`, but don't think that's any ready to be even open as a PR or considered as a fix.

Help needed.

Screen record for `fix/hyprland` branch:

https://github.com/user-attachments/assets/b7a6e8ee-6509-4b41-9050-69a6355fb715

y3owk1n:
To not confuse everything, hyprland does work, the problem is that on hyprland, despite we set an empty input region with "exclusive" keyboard interactivity, the mouse actions are still applied on the overlay itself rather than the window behinds the overlay, unlike sway and niri.

Since everything is performed on the overlay, there's nothing to click or scroll, hence it feels like it's not working at all. To further clarify, cli actions like `neru action left_click` works on hyprland just fine.

y3owk1n:
Found a sort of working way to not changing any code but use neru primitives to make clicking works in hyprland...

```toml
[recursive_grid.hotkeys]
"Shift+L" = ["idle", "action wait_for_mode_exit", "action left_click", "recursive_grid"]
```

What the snippet above does is in sequence:
1. quit the mode
2. wait for it (it's instant)
3. do the left_click on the position
4. relaunch the grid (you can omit this if you don't want to relaunch)

Not the best solution, but if you want to get clicking at least work, this is one of the hacky way. Note that for some reason, scroll is completely non working on hyprland even with just cli calls without overlay...

y3owk1n:
Is this the reason why hyprland doesn't pass through the actions to bottom window?

https://github.com/hyprwm/Hyprland/discussions/11834#discussioncomment-14513354

neel2222:
> Is this the reason why hyprland doesn't pass through the actions to bottom window?
> 
> [hyprwm/Hyprland#11834 (comment)](https://github.com/hyprwm/Hyprland/discussions/11834#discussioncomment-14513354)

If i understood correctly no, neru uses overlay layer.
In issue vaxry mention empty region as way to not focus overlay https://github.com/hyprwm/Hyprland/issues/6059, so maybe that was changed intentionally or it`s a bug.

y3owk1n:
Ya it should work but it don't 😔 I'll continue to do more research when I have time on this ~

y3owk1n:
[...]

PR #731, opened 2026-04-25
feat(linux): add `uinput` scroll support as first detection layer

y3owk1n:
## Description

This PR adds uinput based scrolling support for Wayland compositors. Previously we are solely relying on overlay keyboard interactivity plus empty region input, which works for niri and sway, but not hyprland.

With uinput based scrolling implementation, we can finally support scrolling in hyprland properly. The implementation is very similar to how we made mouse clicking support for hyprland as well.

## Related Issues

Related to #696

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

## General Checklist

## Screenshots / Recordings

https://github.com/user-attachments/assets/a07ff3d4-6d5d-418a-9127-969dfa546de5

## Additional Context

N/A

### DIFF
--- a/internal/app/modes/indicator_polling.go
+++ b/internal/app/modes/indicator_polling.go
@@ -182,6 +182,11 @@ func (h *Handler) startIndicatorPolling(mode domain.Mode) {
 // stopIndicatorPolling stops the indicator polling goroutine and cleans up
 // both mode indicator and sticky modifiers indicator overlays.
 func (h *Handler) stopIndicatorPolling() {
+	// Restore keyboard capture if uinput scroll was active.
+	if m := overlay.Get(); m != nil && eventtap.IsUinputScrollAvailable() {
+		m.SetKeyboardCaptureEnabled(true)
+	}
+
 	// Signal stop first.
 	if h.indicatorStopCh != nil {
 		close(h.indicatorStopCh)
--- a/internal/app/modes/indicator_polling.go
+++ b/internal/app/modes/indicator_polling.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/y3owk1n/neru/internal/core/domain"
+	"github.com/y3owk1n/neru/internal/ui/overlay"
 )
 
 const (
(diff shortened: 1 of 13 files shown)


y3owk1n:
@greptileai i noticed that on wayland niri, the scroll direction configured is scroll_down, but in reality, it scroll upwards, why is that? Please show me the fixes in detail if possible.

y3owk1n:
> The root cause is a sign convention mismatch between the app's internal convention and how libinput (which niri uses) interprets `REL_WHEEL` events.
> 
> **The conventions:**
> 
> * **App convention**: positive `deltaY` = scroll UP, negative `deltaY` = scroll DOWN
> * **wlroots path**: positive Wayland axis = scroll DOWN → the code **negates** deltaY (see `step := -wlrootsScrollStep` in `wlrootsScrollAtCursor`)
> * **uinput/libinput path**: `REL_WHEEL -1` = scroll UP in libinput's convention
> 
> **What happens for `scroll_down`:**
> 
> 1. `scroll_down` → negative `deltaY`
> 2. Current code sends `value = -1` (since `deltaY < 0`)
> 3. niri (via libinput) interprets `REL_WHEEL = -1` as **scroll UP** → wrong!
> 
> The uinput path is missing the same negation the wlroots path applies. The fix in `element_linux.go` for the `deltaY` value block:
> 
> ```
> 			value := -1
> 			if deltaY < 0 {
> 				value = 1
> 			}
> ```
> 
> The horizontal axis (`REL_HWHEEL`) doesn't have this inversion issue — libinput's horizontal convention matches the app convention (positive = right), so the `deltaX` block is correct as-is.
> 
> This negation mirrors exactly what the wlroots fallback path already does with `step := -wlrootsScrollStep`, bringing the uinput first-layer in sync with the rest of the scroll pipeline.

@greptileai i have tried with this fix, but it is still opposite when i actually try it on my machine. both x and y axis behaves the same (opposite). Please go through again and find out where is the bug. Note that i am with uinput.

y3owk1n:
Whatever, since this is a fix to support hyprland scrolling with uinput, i'll give it a pass first until somebody have any issue with it. The other issue that I am seeing is:

With uinput enabled, Niri and hyprland scrolling delta are behaving differently (exactly opposite). The issue now is that I do not know which is the correct one. Studying the specs, the current code should be correct (but i might be wrong). Whoever knows what's going on, feel free to fix it later.

PR #741, opened 2026-05-02
fix: preserve hint action hotkeys with sticky modifiers

snoolord:
## Summary
- Preserve raw hint-mode action hotkeys like `Shift+L` when sticky modifiers strip the key before mode handling.
- Keep stripped-key behavior for normal mode navigation so sticky modifiers do not leak into bindings like `Ctrl+C`.
- Add regression coverage for hint action hotkeys, multi-action bindings, app-specific disabled overrides, and sticky modifier stripping.

## Verification
- `go test ./internal/app/modes ...`
- `just fmt && just lint && just test && just build`

### DIFF
--- a/internal/app/modes/key_dispatch.go
+++ b/internal/app/modes/key_dispatch.go
@@ -14,11 +14,12 @@ import (
 const hotkeySequenceTimeout = 500 * time.Millisecond
 
 const (
-	keyPartCmd    = "cmd"
-	keyPartShift  = "shift"
-	keyPartAlt    = "alt"
-	keyPartCtrl   = "ctrl"
-	keyPartOption = "option"
+	keyPartCmd         = "cmd"
+	keyPartShift       = "shift"
+	keyPartAlt         = "alt"
+	keyPartCtrl        = "ctrl"
+	keyPartOption      = "option"
+	hotkeyActionPrefix = "action "
 )
 
 // HandleKeyPress dispatches key events by current mode.
@@ -63,6 +64,10 @@ func (h *Handler) HandleKeyPress(key string) {
 	// only. Sticky modifiers are for the next action, not Neru's own navigation
 	// keys; using rawKey here would make a sticky Ctrl turn "c" into "Ctrl+c".
 	if rawKey != key {
+		if h.handleRawHintsActionHotkey(rawKey, bundleID) {
+			return
+		}
+
 		if h.handleHotkey(key, bundleID) {
 			return
 		}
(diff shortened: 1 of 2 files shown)


internal/app/modes/key_dispatch.go
```diff
@@ -63,6 +64,10 @@ func (h *Handler) HandleKeyPress(key string) {
 	// only. Sticky modifiers are for the next action, not Neru's own navigation
 	// keys; using rawKey here would make a sticky Ctrl turn "c" into "Ctrl+c".
 	if rawKey != key {
+		if h.handleRawHintsActionHotkey(rawKey, bundleID) {
```
y3owk1n, line 67: Do we really need to handle it here? I think it's better to handle around where we strip the modifiers and keep the handler simple and easy to just handle whatever it gets.

I think we can just make the whole implementation easier, before we strip out the modifiers, we first check if the key that gets in are part of configured key. If it is, we just skip it entirely, no stripping should be done. In this way, we don't have to touch anything on handler and strip functions.

Not sure what do you  think about this? I think its cleaner, lesser code changes and also less likely to introduce any regression.

Quick and dirty implementation to show what i mean:

```go
func (h *Handler) HandleKeyPress(key string) {
	h.mu.Lock()
	defer h.mu.Unlock()

	// Cancel any pending modifier toggle if a non-modifier key is pressed
	// This handles the case where Shift+L is pressed - the modifier tap
	// is canceled when L comes in
	if !strings.HasPrefix(key, modifierTogglePrefix) {
		h.markHeldModifiersUsedInChord()
		h.cancelPendingModifierToggle()
	}

	// Check for modifier toggle keys before any other processing
	if h.handleModifierToggle(key) {
		return
	}

	// Save the raw key before sticky modifier stripping so we can try
	// hotkey matching with the original modifier combo later.
	rawKey := key

	// Resolve the focused app bundle ID once so that both handleHotkey calls
	// (rawKey and stripped key) share the same snapshot.
	var bundleID string
	if h.appState.CurrentMode() == domain.ModeHints && h.config.Hints.HasAppHotkeyOverrides() {
		bundleID = h.focusedBundleID()
	}

	currentModeName := domain.ModeString(h.appState.CurrentMode())
	hotkeys := h.config.HotkeysForModeAndApp(currentModeName, bundleID)

	// Sticky modifiers are also physically posted into macOS so apps can react
	// as if the key is held. Strip those sticky prefixes back out for Neru's own
	// binding resolution so regular mode keys still behave predictably.
	activeMods := h.stickyModifiers()

	isHotkey := false

	if len(hotkeys) > 0 {
		for bindKey := range hotkeys {
			if strings.EqualFold(bindKey, key) {
				h.logger.Info("Key is a hotkey", zap.String("key", key))
				isHotkey = true
				break
			}
		}
	}

	if activeMods != 0 && !isHotkey && !strings.HasPrefix(key, modifierTogglePrefix) {
		key = h.stripStickyModifiersFromKey(key, activeMods)
	}

	// Check for per-mode hotkeys before mode-specific handling.
	// If sticky modifiers were stripped, resolve bindings with the stripped key
	// only. Sticky modifiers are for the next action, not Neru's own navigation
	// keys; using rawKey here would make a sticky Ctrl turn "c" into "Ctrl+c".
	if rawKey != key {
		if h.handleHotkey(key, bundleID) {
			return
		}
	} else if h.handleHotkey(rawKey, bundleID) {
		return
	}

	h.handleModeSpecificKey(key)
}
```

y3owk1n:
By the way, I think this issue was accidentally fixed when I am fixing issue #751 on PR #752 and #754. Feel free to test the nightly build or build from main to test it out. If it actually fixes what this PR tries to do, let's close this.

y3owk1n:
I'll go ahead and close this, as this is already fixed, thanks!

PR #778, opened 2026-05-08
fix(config): replace default launcher hotkeys on rebind

tharropoulos:
## Description

- adjust global hotkey merge behavior so rebinding a built-in launcher action replaces its default launcher instead of adding a second active binding.

- keeps merge-on-top-of-defaults semantics for [hotkeys], but makes single-action rebinds for hints, grid, recursive_grid, and scroll behave like users expect. 
- refactors raw hotkey action parsing into a shared helper and adds integration coverage for the new behavior.

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

## General Checklist

 ## Additional Context

This intentionally changes only top-level global hotkey launcher behavior.  Per-mode hotkeys still use the existing merge semantics, and explicit  __disabled__ handling is unchanged.

### DIFF
--- a/internal/config/service.go
+++ b/internal/config/service.go
@@ -78,6 +78,57 @@ func findNormalizedMapKey[V any](m map[string]V, rawKey string) string {
 	return rawKey
 }
 
+func isBuiltInGlobalModeAction(actions []string) (string, bool) {
+	if len(actions) != 1 {
+		return "", false
+	}
+
+	switch actions[0] {
+	case modeNameHints, modeNameGrid, modeNameRecursiveGrid, modeNameScroll:
+		return actions[0], true
+	default:
+		return "", false
+	}
+}
+
+func removeBindingsForSingleAction(bindings map[string][]string, action string) {
+	for key, existingActions := range bindings {
+		existingAction, ok := isBuiltInGlobalModeAction(existingActions)
+		if ok && existingAction == action {
+			delete(bindings, key)
+		}
+	}
+}
+
+func parseRawHotkeyActions(fieldName string, value any) ([]string, error) {
+	switch val := value.(type) {
+	case string:
+		return []string{val}, nil
+	case []any:
+		actions := make([]string, 0, len(val))
+		for _, actionValue := range val {
+			actionStr, ok := actionValue.(string)
+			if !ok {
+				return nil, derrors.Newf(
+					derrors.CodeInvalidConfig,
+					"%s must be a string or array of strings",
+					fieldName,
+				)
+			}
+
+			actions = append(actions, actionStr)
+		}
+
+		return actions, nil
+	default:
+		return nil, derrors.Newf(
+			derrors.CodeInvalidConfig,
+			"%s must be a string or array of strings",
+			fieldName,
+		)
+	}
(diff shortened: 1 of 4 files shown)


internal/config/service.go
```diff
@@ -78,6 +78,57 @@ func findNormalizedMapKey[V any](m map[string]V, rawKey string) string {
...
 }
 
+func isBuiltInGlobalModeAction(actions []string) (string, bool) {
+	if len(actions) != 1 {
+		return "", false
+	}
+
+	switch actions[0] {
+	case modeNameHints, modeNameGrid, modeNameRecursiveGrid, modeNameScroll:
+		return actions[0], true
```
y3owk1n: There might be cases like my config for example that does `"Ctrl+F" = "recursive_grid --cursor-selection-mode hold"`. This might be a gotcha for a user that started with `Ctrl+F = "recursive_grid`, and we replaced the default, but later made this binding to a more advance version, and suddenly the default binding is not disabled.

I think we should cover this case too, and also another test case for this scenario would be nice.

internal/config/service.go
```diff
@@ -78,6 +78,57 @@ func findNormalizedMapKey[V any](m map[string]V, rawKey string) string {
 	return rawKey
 }
 
+func isBuiltInGlobalModeAction(actions []string) (string, bool) {
```
y3owk1n, line 83: Let's have some annotation to briefly explain about what this function does for consistency, same goes to other functions down below.

y3owk1n:
Thanks for the PR, overall very solid and straight forward, it definitely helps a lot for new users that are just testing the program out, by confining the scope to only built-in keybindings, the 4 defaults, the mentality of it changes and it make a lot of sense now.

This had been bugging me for a while, and I am unable to get the right method, probably because i was thinking too big and wanted everything to be automatically dedupe, but it's not easy.

I had just left some comment, but mainly no big deal. Once we resolve them, i'll let greptile to run a bug check, and if it's good, we are good to go.

y3owk1n (changes requested):

internal/config/service.go
tharropoulos: added them on 774fb0b

internal/config/service.go
tharropoulos: addressed in [9d8c37e](https://github.com/y3owk1n/neru/pull/778/commits/9d8c37e35f81fd39dbf04d808f814747b99f859c)

y3owk1n: Looks good, thanks!

internal/config/service.go
y3owk1n: perfect ~

y3owk1n:
@greptileai

y3owk1n (approved):

y3owk1n:
We're good, just to leave a note here for future potential, with the idea of disabling defaults, we might be able to also detect if certain mode is disabled, then we can go ahead and disable the default for that mode too. Not sure if it's a good idea.

PR #782, opened 2026-05-08
feat(linux): detect session dark mode via freedesktop appearance portal

T-Py-T:
## Description

Implements the long-standing Linux `IsDarkMode` stub (`return false`) by querying the xdg-desktop-portal Settings interface for `org.freedesktop.appearance` `color-scheme`, with a `~/.config/kdeglobals` `[General] ColorScheme` fallback for vanilla KDE Plasma installs that don't have `xdg-desktop-portal-kde` and where the gtk-portal returns nothing useful.

The Linux `SystemAdapter` now live-probes on every `Capabilities()` call and embeds the current state in `DarkModeDetection.Detail`, so `neru doctor` shows the actual dark-mode value the daemon is seeing — not just whether the code path exists. When no source is reachable the capability degrades gracefully to `stub` with a fix-it hint pointing at the right packages to install.

### Implementation notes

- Source ordering: `xdg-desktop-portal` first (works on GNOME, KDE-with-portal-kde, and any wlroots compositor with portal-gtk fallback), then `~/.config/kdeglobals` (covers vanilla KDE installs without `xdg-desktop-portal-kde`).
- The portal's `Settings.Read` returns a `variant variant uint32`; `busctl` formats that as e.g. `v v u 1`. We take the trailing token.
- **`busctl --quiet` suppresses the method-call return value** (not just bus chatter) — the helper deliberately omits it. There's a comment in `readPortalColorScheme` calling this out so future contributors don't "clean it up" and silently break detection.
- `darkModeCapability` keeps the surface honest: only reports `supported` when an actual source produced a value. Otherwise `stub` with a real fix-it hint.

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (`//go:build linux` on the implementation and the test)
- [x] No darwin imports from untagged (shared) code — only `internal/app/ipc_info.go` and `internal/cli/cliutil/util.go` are shared, and both treat the detail field generically (other adapters that don't populate `Detail` simply don't surface a `Dark Mode:` line).
- [x] Stub implementations added for other platforms — Darwin already has `IsDarkMode` via Cocoa; Windows still returns its existing stub.

## General Checklist

- [x] Code formatted
- [x] Linters pass (`golangci-lint run ./...` → `0 issues.`)
- [x] Tests pass (`go test ./...`)
- [x] Build succeeds (`just build-linux arm64`, also `go build ./...` on darwin/windows tags)
- [x] Tests added — see `system_linux_dark_mode_test.go`:
  - `TestScanINIValueFindsKeyInCorrectSection` — section/key isolation
  - `TestScanINIValueReturnsEmptyWhenSectionOrKeyMissing` — three negative cases
  - `TestDarkModeCapabilitySupportedStates` — dark/light/no-preference for both portal and kdeglobals sources
  - `TestDarkModeCapabilityDowngradesToStubWhenUnreachable` — verifies the fix-it hint
  - `TestReadKDEColorSchemeIsDarkWhenSchemeNameContainsDark` — case-insensitive `*Dark` substring inference
- [x] Commit message follows conventional commits

## Verification

End-to-end verified against three live VMs (all UTM aarch64) with all three color-scheme states:

| VM | Desktop | Portal value | `IsDarkMode()` | `doctor` `Dark Mode:` |
|----|---------|--------------|----------------|------------------------|
| Ubuntu 26.04 | GNOME (Wayland) | `v v u 1` | true | `current state: dark (source=xdg-portal)` |
| Fedora 44 KDE | Plasma 6 (Wayland) | `v v u 1` | true | `current state: dark (source=xdg-portal)` |
| Fedora 44 GNOME | GNOME 47 (Wayland) | `v v u 1` | true | `current state: dark (source=xdg-portal)` |

Live toggling with `gsettings set org.gnome.desktop.interface color-scheme prefer-light/prefer-dark/default` updates the `doctor` output on the next call without restarting the daemon.

## Additional Context

The runtime probe runs on every `Capabilities()` call, which today only happens when `neru status` or `neru doctor` is invoked (cheap: sub-100ms `busctl` call on a healthy session bus, capped at 250ms via context timeout). If a future caller starts hitting this in a hot path, caching with invalidation on the portal's `SettingChanged` signal would be the natural follow-up — but that's premature today.

### DIFF
--- a/internal/app/ipc_info.go
+++ b/internal/app/ipc_info.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"go.uber.org/zap"
@@ -253,6 +254,14 @@ func (h *IPCControllerInfo) handleHealth(ctx context.Context, _ ipc.Command) ipc
 	}
 
 	for key, value := range capabilities {
+		// Skip informational sibling fields (e.g. dark_mode_detection_detail).
+		// These are surfaced through the metadata header in `neru doctor`, not
+		// as component health rows, and their values are free-form prose that
+		// doesn't fit the supported/stub/headless/ok vocabulary.
+		if strings.HasSuffix(key, "_detail") {
+			continue
+		}
+
 		status, _ := value.(string)
 
 		components["capability."+key] = status
@@ -351,7 +360,7 @@ func capabilitiesMap(capabilities ports.PlatformCapabilities) map[string]any {
 		return map[string]any{}
 	}
 
-	return map[string]any{
+	out := map[string]any{
 		"platform":            capabilities.Platform,
 		"process":             capabilityString(capabilities.Process),
 		"screen":              capabilityString(capabilities.Screen),
@@ -364,6 +373,17 @@ func capabilitiesMap(capabilities ports.PlatformCapabilities) map[string]any {
 		"app_watcher":         capabilityString(capabilities.AppWatcher),
 		"dark_mode_detection": capabilityString(capabilities.DarkModeDetection),
 	}
+
+	// Surface dark-mode Detail as a sibling field so `neru doctor` can render
+	// the current live state (e.g. "current state: dark (source=xdg-portal)")
+	// without having to expand FeatureCapability into its own structured map.
+	// Adapters that don't populate Detail (currently Darwin/Windows) just won't
+	// include the key.
+	if detail := capabilities.DarkModeDetection.Detail; detail != "" {
+		out["dark_mode_detection_detail"] = detail
+	}
+
+	return out
 }
 
 func profileMap(profile platform.Profile) map[string]any {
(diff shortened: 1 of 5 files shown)


T-Py-T:
CI for this PR is waiting on **Approve and run workflows** (first-time contributor fork). Local verification on the PR commit: `golangci-lint run ./...` → 0 issues; `go test ./...` → green on Linux (Ubuntu VM) and Darwin packages. Fork `main` is updated with the same change as a single squash commit (`e79748b`) for anyone who needs a merged line before upstream merges.

y3owk1n:
Oh, i was about to check out this PR, what's wrong?

T-Py-T:
> Oh, i was about to check out this PR, what's wrong?

I accidentally made the PR against yours instead of mine. I need some more testing and package deployment testing. I would also like to do some regression testing on X11. 

I spent a decent amount of time setting up the different VMs and findings the items that output. Im still working through how to test cleanly that the fix does whats needed and passes the CI test ( i see that it failed).

y3owk1n:
I see, no worries ~ take your time. If you want, i dont think you need to make it fully complete first. We can do it phase by phase too. Say the first implementation only supports wayland, and the rest are no-op. And then slowly port to other distros too. Dont stress yourself up.

PR #817, opened 2026-05-17
feat(hotkey): repeat global hotkey actions while held automatically

### LINKED ISSUE #816: Repeat a global hotkey action while the key is held
### Pre-flight checks

- [x] I have searched existing issues and discussions and this has not been requested yet

### Category

Mouse actions (click, drag, scroll)

### Target platform

macOS

### Problem or use case

For quick scrolling, the natural workflow today is to enter scroll mode (`Primary+Shift+S`), hold `j` or `k`, then `Escape` out. Inside the mode the keys repeat as expected.

For short, "I just want to nudge this down a bit" scrolls, the mode entry/exit adds friction. The natural alternative is to bind a scroll action directly to a global hotkey:

```toml
[hotkeys]
"Primary+Alt+J" = "action scroll_down"
"Primary+Alt+K" = "action scroll_up"
```

A single tap works fine. But holding the hotkey down only fires the action once — the action doesn't repeat while held. From the user's point of view this breaks the "press-and-hold to keep doing the thing" intuition that the in-mode `j`/`k` already follows.

My use case: a fluent press-and-hold of a global hotkey that scrolls continuously until released, without having to step in and out of scroll mode for every short scroll.

### Proposed solution

A way to mark a global hotkey binding so that, while the hotkey is held, the bound action keeps running, and stops on release.

I don't want to over-prescribe the config shape — anything that lets the user opt in per binding would work. Roughly:

- Tap (release before some threshold): fire the action once, as today.
- Hold past the threshold: keep firing the action until the user releases the hotkey.

I'm not familiar enough with neru's internals to say what's cheapest to implement. I noticed `neru doctor` reports both a Carbon hotkey listener and a Quartz event tap, so the keyup side of the chord seems observable without extra permissions, but I may be missing something.

### Alternatives considered

**Scroll mode.** What I use today. Works perfectly for longer scroll sessions. Just feels heavy for the short ones.

**Multiple actions chained in the binding.** I tried something like:

```toml
"Primary+Alt+J" = ["action scroll_down", "action scroll_down", "action scroll_down"]
```

It fires N scrolls per tap, but the chain is fire-and-forget — it keeps running even if I release the key partway, and I have to guess the right length. Not really equivalent to "scroll while held".

**`--action` and `--repeat` flags.** Closest existing features I found. `--action` lets a mode entry execute the action and exit (`#484` / `auto_exit_actions`). `--repeat` (#637) re-activates the mode after the action. Both still require entering a mode and operate on the post-action transition, which is conceptually different from what this asks for (no mode at all, repeat triggered by physical hold).

**Karabiner-Elements externally.** I went down this rabbit hole. Karabiner's `to_if_held_down` fires its action array exactly once per press and has no documented loop-while-held primitive. A maintainer confirms in [pqrs-org/Karabiner-Elements#1293](https://github.com/pqrs-org/Karabiner-Elements/issues/1293) that "to while held down" is intentionally not how Karabiner works, and a CLI keypress feature request ([#2966](https://github.com/pqrs-org/Karabiner-Elements/issues/2966)) was closed as not planned. The closest I got was a finite burst of ~60 synthetic events with `hold_down_milliseconds` between them, which works but isn't interruptible on release.

**Hammerspoon.** Possible in principle (event tap + timer that posts synthetic key events until keyup). I haven't been able to confirm whether Hammerspoon-emitted events on recent macOS reach a Carbon `RegisterEventHotKey` handler reliably — that path has been reported as flaky in other contexts. Either way it's an external dependency for behavior that fits cleanly inside the app.

I'm raising this here because none of the external workarounds gives the same guarantee as having it live in neru itself.

### Screenshots / mockups

_No response_

### Contribution

y3owk1n:
## Description

This PR adds automatic hold-repeat support for repeatable global hotkey actions on macOS, but with certain restrictions that we think that it is worth repeating:

- The repeatable action needs to be one of `action scroll_xxx`, `action page_xxx`, or `action move_mouse_relative ...`
- The repeatable action is binded to global hotkey as a string

```toml
# these repeats...
[hotkeys]
"Primary+Alt+J" = "action scroll_down"
"Primary+Alt+K" = "action scroll_up"
"Primary+Alt+Down" = "action page_down"
"Primary+Alt+L" = "action move_mouse_relative --dx=10 --dy=0"

# these DOES NOT repeat....
"Primary+Alt+J" = ["action scroll_down", "hints", "action scroll_down"]
"Primary+Alt+J" = ["action scroll_down", "action scroll_down"]
"Primary+Alt+J" = "hints"
"Primary+Alt+J" = "action left_click"
"Primary+Alt+J" = "exec echo hi"
```

This change does not introduce any breaking changes nor config changes, it should just work as it is.

## Related Issues

Closes #816

## Target Platform

- [x] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

## General Checklist

- [ ] Documentation updated (if applicable)

## Screenshots / Recordings

N/A

## Additional Context

The implementation uses Carbon hotkey release events instead of relying on the Quartz event tap. That keeps idle global hotkey repeat tied to the same backend that already owns global hotkey activation.

Repeat support is intentionally conservative and only applies to single bindings whose action naturally behaves like native key repeat. New actions that are expected to repeat needs to be explicitly added into the list.

### DIFF
--- a/internal/app/app_config.go
+++ b/internal/app/app_config.go
@@ -41,6 +41,7 @@ func (a *App) prepareForConfigUpdate() {
 
 	if a.appState.HotkeysRegistered() {
 		a.logger.Info("Unregistering current hotkeys before reload")
+		a.stopAllHotkeyRepeats()
 		a.hotkeyManager.UnregisterAll()
 		a.appState.SetHotkeysRegistered(false)
 	}
--- a/internal/app/app_initialization_steps.go
+++ b/internal/app/app_initialization_steps.go
@@ -451,6 +451,7 @@ func initializeShutdownChannel(app *App) {
 func cleanupInfrastructure(app *App) {
 	// Clean up hotkey service
 	if app.hotkeyManager != nil {
+		app.stopAllHotkeyRepeats()
 		app.hotkeyManager.UnregisterAll()
 		app.hotkeyManager = nil
 	}
--- a/internal/app/app_types.go
+++ b/internal/app/app_types.go
@@ -66,6 +66,9 @@ type App struct {
 	// concurrent writers (theme change observer, IPC config reload, systray reload).
 	configMu sync.RWMutex
 
+	hotkeyRepeatMu      sync.Mutex
+	hotkeyRepeatCancels map[string]context.CancelFunc
+
 	// New Architecture Services
 	hintService            *services.HintService
 	gridService            *services.GridService
--- a/internal/app/hotkeys.go
+++ b/internal/app/hotkeys.go
@@ -140,17 +140,25 @@ func (a *App) dispatchHotkeyActions(key string, actions []string) {
 }
 
 func (a *App) startHotkeyRepeat(key string, actions []string) {
-	a.stopHotkeyRepeat(key)
-
 	ctx, cancel := context.WithCancel(context.Background())
 
 	a.hotkeyRepeatMu.Lock()
+
+	oldCancel := a.hotkeyRepeatCancels[key]
+	if oldCancel != nil {
+		delete(a.hotkeyRepeatCancels, key)
+	}
+
 	if a.hotkeyRepeatCancels == nil {
 		a.hotkeyRepeatCancels = make(map[string]context.CancelFunc)
 	}
 
 	a.hotkeyRepeatCancels[key] = cancel
 	a.hotkeyRepeatMu.Unlock()
+
+	if oldCancel != nil {
+		oldCancel()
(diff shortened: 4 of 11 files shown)


y3owk1n:
@greptileai

y3owk1n:
@greptileai

PR #882, opened 2026-06-01
feat(default-config): promote 3x3 recursive grid to default grid

y9san9:
As discussed earlier, 2x2 feels "depressing" when you see it for the
first time. And labels feel like a more advanced solution for the
problem. Tho @y3owk1n and many experienced users switch from them for
recursive grid.

This is a step in the direction that will help people to love recursive
grid ~~and forget labels~~.

Keys Justification:

I picked `rtyfghvbn` keys because they are physically a square in the
middle of keyboard. Placed just right, so you can use both of your hands
to pick them.

In my opinion it is the most newbie-friendly approach. But it's not the
most efficient for sure. For experienced user it is more preferrable to
use keys like `asdf jkl;`. Because they are all on homerow! And you
never need to move anywhere to click.

But it needs additional explanation and probably not the best fit for
newbies.

---

On animation part: I can revert it if you don't agree with that. But I
think again that it's more user-friendly and will give some joy to a
modern soy dev (I'm using this myself, lol). Feel free to ask me to
revert this if you don't agree.

### DIFF
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -186,16 +186,16 @@ border_width = 1
 # See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#recursive_grid
 [recursive_grid]
 enabled = true
-grid_cols = 2                               # Number of columns (>= 1)
-grid_rows = 2                               # Number of rows (>= 1)
-keys = "uijk"                               # Cell selection keys (must be grid_cols * grid_rows chars)
+grid_cols = 3                               # Number of columns (>= 1)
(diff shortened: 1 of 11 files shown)


y3owk1n:
@y9san9 i have no strong opinion on the defaults tho, seems reasonable. i'll take over a bit to just update some documentation along side this change, so that we don't have to do 2 PR for this.

y9san9:
sure!

y3owk1n:
Done! once CI passes, we are good to go.

PR #892, opened 2026-06-04
feat(linux): detect dark mode via xdg portal + kdeglobals

T-Py-T:
Scoped down per your review to Linux dark mode detection only.

What's new:
- Reads the color scheme from xdg-desktop-portal (org.freedesktop.appearance via Settings.Read), falling back to ~/.config/kdeglobals.
- Surfaces it in `neru doctor` as a "Dark Mode:" line (state + source).
- Flips the Linux DarkModeDetection capability from stub to supported; the SystemAdapter overrides it with the live value on each call.
- One line in docs/LINUX_SETUP.md.

The rest of the Linux capability work this branch originally carried (wlroots/KDE input + overlay) is reverted back to main. Those stay off until the libei / RemoteDesktop portal path is done and working, which I'll send as a separate PR (KWin won't implement zwlr_virtual_pointer_v1; WONTFIX, bug 518354).

Review comments: dropped logger export_test.go (the dark mode test uses package linux with //nolint:testpackage); no validation-linux docs; default-config.toml untouched; the dark mode code is universal so it stays in system_linux_common.go (portal covers GNOME/KDE/wlroots/X11, kdeglobals is a KDE-only fallback). Logging isn't in this PR anymore since it already landed on main in #908.

Testing: built on a clean Fedora 44 KDE VM (Plasma 6.6.4, arm64, CGO on); unit tests pass; a Breeze Dark session returns 1 and parses correctly. A full `neru doctor` on KDE needs the KDE backend that isn't on main yet, so that end-to-end check comes with the libei work.

### DIFF
--- a/internal/core/infra/platform/linux/system_linux_common.go
+++ b/internal/core/infra/platform/linux/system_linux_common.go
@@ -4,10 +4,16 @@
 package linux
 
 import (
+	"bufio"
 	"context"
 	"image"
+	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
 
 	derrors "github.com/y3owk1n/neru/internal/core/errors"
 	"github.com/y3owk1n/neru/internal/core/ports"
@@ -34,12 +40,23 @@ func (s *SystemAdapter) Health(ctx context.Context) error {
 }
 
 // Capabilities returns the current Linux capability surface.
+//
+// The dark-mode capability is live-probed: if a working source can be reached
+// the Detail field carries the current state ("dark" / "light" / "no
+// preference") plus the source name; if none of the sources work the status
+// is downgraded to stub with a fix-it hint. This is more useful than a
+// static "supported" claim because the user's actual question when running
+// `neru doctor` is "is dark mode being detected right now?", not "does the
+// code path exist on Linux?".
 func (s *SystemAdapter) Capabilities() ports.PlatformCapabilities {
 	capabilities := ports.LinuxCapabilities()
 
 	if s.backend != "" {
 		capabilities.Platform = "linux/" + s.backend
 	}
+
+	value, source, ok := darkModePreference()
+	capabilities.DarkModeDetection = darkModeCapability(value, source, ok)
 
 	return capabilities
 }
@@ -218,10 +235,12 @@ func (s *SystemAdapter) CursorPosition(ctx context.Context) (image.Point, error)
 	)
 }
 
-// IsDarkMode returns true if Linux dark mode is currently active.
-// TODO(linux): implement using org.freedesktop.appearance color-scheme D-Bus property.
+// IsDarkMode returns true if Linux dark mode is currently active. See
+// darkModePreference for source ordering and semantics.
 func (s *SystemAdapter) IsDarkMode() bool {
-	return false
+	value, _, ok := darkModePreference()
+
+	return ok && value == colorSchemeDark
 }
 
 // CheckPermissions verifies accessibility permissions on Linux.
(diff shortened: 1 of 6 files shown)


y3owk1n:
@T-Py-T thanks! Could you possible provide a screen record on KDE that briefly goes through what has implemented to ensure everything working? This gives me more confidence to merge this later when i start looking into the code.

T-Py-T:
> @T-Py-T thanks! Could you possible provide a screen record on KDE that briefly goes through what has implemented to ensure everything working? This gives me more confidence to merge this later when i start looking into the code.

Of course. I would have to look at how to do that. I will upload tomorrow, as its late here. Ill record the KDE VM that I have on UTM. The hints is VERY crude, but working, the rest look ok. Thanks!!!!

y3owk1n:
> > @T-Py-T thanks! Could you possible provide a screen record on KDE that briefly goes through what has implemented to ensure everything working? This gives me more confidence to merge this later when i start looking into the code.
> 
> Of course. Im would have to look at how to do that. I will upload tomorrow, as its late here. Ill record the KDE VM that I have on UTM. The hints is VERY crude, but working, the rest look ok. Thanks!!!!

since you use UTM, you can just do screen record on the macOS layer, and then do the thing in the UTM window, that's how i do it instead of spending time to figure out how to record on linux distros.

Thank you for all of these!

T-Py-T:
https://github.com/user-attachments/assets/97cfa159-0a01-46ff-8063-9263941ae6d8

> > > @T-Py-T thanks! Could you possible provide a screen record on KDE that briefly goes through what has implemented to ensure everything working? This gives me more confidence to merge this later when i start looking into the code.
> > 
> > 
> > Of course. Im would have to look at how to do that. I will upload tomorrow, as its late here. Ill record the KDE VM that I have on UTM. The hints is VERY crude, but working, the rest look ok. Thanks!!!!
> 
> since you use UTM, you can just do screen record on the macOS layer, and then do the thing in the UTM window, that's how i do it instead of spending time to figure out how to record on linux distros.
> 
> Thank you for all of these!

There is a link above in the description for the Fedora 44 KDE Iso (arm64) that I used to create the UTM VM

@y3owk1n  Ok. Video uploaded!!!

configs/default-config.toml
```diff
@@ -355,7 +355,7 @@ enabled = true                              # Show menu bar icon (false = headle
 # See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#logging
 [logging]
 log_level = "info"                          # debug, info, warn, error
-log_file = ""                               # Custom log path (empty = default location)
+log_file = ""                               # Empty: platform default (macOS: ~/Library/Logs/neru/app.log, Linux: ~/.local/state/neru/log/app.log, Windows: %LOCALAPPDATA%/neru/log/app.log)
```
y3owk1n: i think we can remain the old comments and put this info into `docs/CONFIGURATION.md` instead.

docs/validation-linux-amd64.md
y3owk1n, line 1: I think we probably dont need this file at all, instead we should keep `docs/LINUX_SETUP.md` up to date with relevant KDE information, and that's more than enough.

docs/validation-linux-arm64.md
y3owk1n, line 1: same goes to this file.

internal/core/infra/logger/export_test.go
y3owk1n, line 1: I think this file is not necessary, this export is mainly to allow `internal/core/infra/logger/default_log_path_test.go` access the API. I don't mind to just add `//nolint:testpackage` to `internal/core/infra/logger/default_log_path_test.go` and do `package logger` directly.

internal/core/infra/platform/linux/system_linux_common.go
```diff
@@ -248,5 +268,203 @@ func (s *SystemAdapter) ShowAlert(ctx context.Context, title, message string) er
...
+	colorSchemeDark         = 1
+	colorSchemeLight        = 2
+)
+
+// darkModePortalTimeout caps the busctl call. The portal call is normally
+// sub-millisecond on a healthy session bus; 250ms gives us ample margin
+// without blocking `neru doctor` if the portal is wedged.
+const darkModePortalTimeout = 250 * time.Millisecond
+
+// darkModePreference returns the active freedesktop color-scheme preference.
```
y3owk1n, line 297: Question, is these dark mode implementation specific to KDE only or specific to Wayland only or it is compatible for all distros?

If it's specific to KDE only, i would put these into a separated maybe `system_linux_wayland_kde.go`, and stub the rest.

If it's specific to Wayland only, i would want it to be in `system_linux_wayland.go`, and stub the rest.

If it's universal, then this is all good.

y3owk1n:
@T-Py-T , I have just done a quick glance through, and appreciate if you could take a look with it. Other than that, i think the implementations are truly minimal, and not a lot to comment about.

Overall amazing work! I have yet to be able to actually test it on a proper KDE machine, but looking at the screen record, i think i can trust it. What i would do instead later is to ensure it doesn't break anything (looking at the code, it should not break anything tho, but just to be sure) on hyprland and niri, which i already have it setup on my VMs.

y3owk1n (changes requested):

internal/core/infra/platform/linux/system_linux_common.go
y3owk1n: @T-Py-T I think this is the one left to address or confirm, the rest are looking good to me.

T-Py-T:
@y3owk1n 

brew install - I retested the bare VM attempt to flush out the seeting changes that I made. I think brew install would need to know this for installing on that OS/desktop.  

This ran into a snag: pulling the latest Fedora KDE ISO (Fedora 44, Plasma 6.6.4 / KWin 6.6.4), we hit a bump reproducing the KDE path on a clean VM. Newer KWin doesn't implement zwlr_virtual_pointer_v1 and KDE won't add it (bug 518354, WONTFIX), so the wlroots input path can't work on Plasma.

That made me question whether relying on the wlr protocols is the right long-term bet for KDE. The compositor-blessed path is libei via the RemoteDesktop portal, which is also what GNOME/COSMIC use — so it's likely the better foundation for non-wlroots support.

Want your read first. Should this get put back into draft? I dont know if I can re-use this if its wlroots based.

I would also be interested in getting a copy of your Niri/hyperland VM that you are using. 

Here is what I currently have (I had a hard time finding an iso with hyperland/niri desktop)

image

y3owk1n:
> This ran into a snag: pulling the latest Fedora KDE ISO (Fedora 44, Plasma 6.6.4 / KWin 6.6.4), we hit a bump reproducing the KDE path on a clean VM. Newer KWin doesn't implement zwlr_virtual_pointer_v1 and KDE won't add it (bug 518354, WONTFIX), so the wlroots input path can't work on Plasma.
> 
> That made me question whether relying on the wlr protocols is the right long-term bet for KDE. The compositor-blessed path is libei via the RemoteDesktop portal, which is also what GNOME/COSMIC use — so it's likely the better foundation for non-wlroots support.

Thanks for the finding on this, can i safely assume that for future KDE releases, wlroots implementation will not work on it right? Since libei seems to be another alternative that gnome or cosmic uses, i think we should have another path of implementation for libei, but we'll need a proper detection on compositors or whatever that linux called them, and route to the right implementation. I do not understand much on on what you said, but i trust it for now.

Potentially the route should be:

- detect wayland and sessions or anything that differs
- if wlroot supported (say niri, hyprland, etc), use the wlroot implementation
- if on gnome, cosmic, kde, use the libei implementation

I hope i understand the whole situation correctly.
 
> Want your read first. Should this get put back into draft? I dont know if I can re-use this if its wlroots based.

In this case, i think this PR doesn't bring lots of value at this moment, but i do think that there are part of this PR is useful, particularly on the log path for cross platform. I think this could be a new small PR.

This PR can be either move to draft or close depending on if you would like to continue iterate on it. But if we are moving to a full libei implementation, i think we can just close this PR instead and work on a more proper solution.

> I would also be interested in getting a copy of your Niri/hyperland VM that you are using.

My setup is a bit manual, both of my niri and hyprland are using the same fedora wayland ISO.

- For niri, i just boot it up, and install Niri directly following this guide on fedora (https://niri-wm.github.io/niri/Getting-Started.html). I did not do any customisation at all.
- For hyprland, it's a bit more involved, and i actually installed it via this script (https://github.com/JaKooLit/Fedora-Hyprland/)

internal/core/infra/platform/linux/system_linux_common.go
T-Py-T: I have addressed it separately and tried to remove the wayland components. Please review and let me know

y3owk1n:
@T-Py-T i walk through the code and i dont have any comments as it seems straight forward for the most part. I'll pass to greptile to check for some potential bug next. Once those are resolved, we'll ensure all CI passes and we are good to merge.

y3owk1n:
@greptileai

T-Py-T:
@y3owk1n  I cant request review from greptile directly. I made the additions that were recommnded.

y3owk1n:
> @y3owk1n I cant request review from greptile directly. I made the additions that were recommnded.

Thanks for the changes! Looks good to me. Let's do 1 more pass on greptile.

y3owk1n:
@greptile

y3owk1n (dismissed):
LGTM

y3owk1n:
@T-Py-T can you ensure lint passes as well?

T-Py-T:
> @T-Py-T can you ensure lint passes as well?

I just pushed a change for the linter faliures. This should be what you need.

y3owk1n:
@T-Py-T thanks for the effort! We are good to merge.

y3owk1n (approved):

PR #947, opened 2026-06-26
fix(windows): consume non-modifier key-down events to prevent passthrough to apps

### LINKED ISSUE #937: Mode inputs should not passthrough to application
Follow up of PR #932 

When a mode is activated in windows, key presses in the mode should not pass through to the app behind it.

Repro:
1. Open an app (focus on some input)
2. Activate any neru mode
3. Press mode keys

Expected:
- It should do the mode actions like navigating through grid (which works now)
- The key presses should not pass to the input or app behind

y3owk1n:
## Description

This PR fixes a bug where in a mode, key presses works but passthrough to the behind app at the same time.

The fix adds a bool return to the hook callback — when it returns true, the hook procedure returns 1 (non-zero) instead of calling CallNextHookEx, which tells Windows to consume the key. The Windows event tap now:

- Consumes non-modifier key-down events (single characters, Shift+letters, named keys like Return/Space/arrows)
- Passes through modifier-only keys (Ctrl, Alt, Shift, Win), key-up events, and system modifier combos (Ctrl+C, Alt+Tab, Win+D) — so system shortcuts still work while a mode is active

This matches macOS behavior where all non-modifier key-down events are consumed by the event tap (return NULL).

## Related Issues

Fixes #937

## Target Platform

- [ ] macOS
- [ ] Linux
- [x] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

## General Checklist

- [ ] Documentation updated (if applicable)

## Screenshots / Recordings

N/A

## Additional Context

N/A

### DIFF
--- a/internal/core/infra/eventtap/eventtap_windows.go
+++ b/internal/core/infra/eventtap/eventtap_windows.go
@@ -168,9 +168,9 @@ func IsUinputScrollAvailable() bool {
 	return false
 }
 
-func (et *EventTap) handleKey(key string, isUp bool) {
+func (et *EventTap) handleKey(key string, isUp bool) bool {
 	if key == "" {
-		return
+		return false
 	}
 
 	if mod := normalizeWindowsModifier(key); mod != "" {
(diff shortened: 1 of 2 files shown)


y3owk1n:
@greptileai

PR #966, opened 2026-07-02
fix(windows): correct active monitor detection

raymond-w-ko:
## Summary

- pass the Windows POINT argument to MonitorFromPoint by value instead of passing a pointer
- add Windows coverage for packed monitor coordinates, including left-of-primary negative X layouts

## Testing

- go test ./internal/core/infra/platform/windows
- gofmt -w internal/core/infra/platform/windows/win32.go internal/core/infra/platform/windows/win32_test.go
- git diff --check
- just --shell C:\opt\Git\bin\sh.exe --shell-arg -cu build

## Notes

- just fmt and just lint could not run in this Windows Git Bash environment because golangci-lint was not available on PATH.
- go test ./... currently fails in internal/cli when a local Neru daemon is already running; focused Windows platform tests pass.

### DIFF
--- a/internal/core/infra/platform/windows/win32.go
+++ b/internal/core/infra/platform/windows/win32.go
@@ -236,10 +236,8 @@ func enumerateMonitors() ([]displayMonitor, error) {
 func activeScreenBounds() (image.Rectangle, error) {
 	cursor, err := cursorPosition()
 	if err == nil {
-		pt := winPoint{x: int32(cursor.X), y: int32(cursor.Y)}
-
 		ret, _, pointErr := procMonitorFromPoint.Call(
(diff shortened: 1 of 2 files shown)


raymond-w-ko:
I have compiled and personally tested the change. before it was always stuck to my rightmost monitor, and now it works across all 3 of my monitors.

y3owk1n (approved):
Straight to the point, looks good.

y3owk1n:
@greptileai

y3owk1n:
> I have compiled and personally tested the change. before it was always stuck to my rightmost monitor, and now it works across all 3 of my monitors.

Everything looks good, i will merge it now, thanks for the PR and support!

PR #993, opened 2026-07-08
feat(hotkeys): let per-mode hotkeys override a global hotkey on the same key

gabrielecirulli:
Disclaimer: This code was written by Claude. I managed and steered the LLM towards a solution I thought sensible.

# Rationale

I wanted to cycle through all the modes from a single key. The idea was that I'd press `Primary+Ctrl+F` to open hints, and pressing it again would take me to the next mode, cycling through hints, recursive grid, grid, and scroll before looping back to hints. The obvious way to set that up is to have the global key open hints, and then have the same key inside each mode point to the next one in the loop.

```toml
[hotkeys]
"Primary+Ctrl+F" = "hints"

[hints.hotkeys]
"Primary+Ctrl+F" = "recursive_grid"
[recursive_grid.hotkeys]
"Primary+Ctrl+F" = "grid"
[grid.hotkeys]
"Primary+Ctrl+F" = "scroll"
[scroll.hotkeys]
"Primary+Ctrl+F" = "hints"
```

That didn't work. As soon as I was inside a mode, pressing the key just reopened hints instead of moving me forward, so the cycle never got past the first step.

# Issue

Global hotkeys and per-mode hotkeys go through two separate event taps, and the global one catches the key and fires its action before the mode's tap ever receives the event. That means a global binding always wins over a per-mode binding on the same key, and there was no way to make a global key do something different while you were inside a mode.

# Changes

This PR makes it so that when a global hotkey fires, it first checks whether the active mode binds that same key. If it does, the mode's binding runs instead. If it doesn't, the global action runs exactly like it used to. Global keys still work from idle, and they still let you jump straight to another mode from inside one, but a mode can now override a key for as long as it's active. That's enough to make the cycle work.

# Repeat keys

I ran into a second version of the same issue while I was making this change. Some global keys are meant to repeat while you hold them, such as scroll, page, and cursor nudge when held-key repeat is enabled, and those go through a different registration path that also skips the override. That path checks whether a key repeats, once during startup, based on the global action. I moved the decision to press time so it's based on whichever binding actually wins, which makes the override work on that path too.

# Notes

There's also a small refactor that pulls a repeated per-app check into a shared helper.

You can still jump to another mode using its global hotkey from inside a mode, because when the current mode doesn't bind that key the global action runs as before. 

Windows and Linux are unaffected, since their hotkey backend can't report key releases, so they never had hold-to-repeat and their path already resolved the override.

For tests I covered the override lookup across the cases that matter (mode binds the key, mode doesn't, idle, and a key that only lives in another mode's table), the press-time repeat decision for every override-and-global combination, and the registration wiring where a press starts the repeat and the release stops it.

# Potential improvements

Key matching doesn't sort modifiers, so `Cmd+Ctrl+F` and `Ctrl+Cmd+F` are treated as different keys for the override. That's a pre-existing quirk in the shared normalize function, and fixing it felt like a separate change.

### DIFF
--- a/internal/app/hotkeys.go
+++ b/internal/app/hotkeys.go
@@ -153,7 +153,12 @@ func (a *App) dispatchModeAwareHeldHotkey(key string, globalActions []string) {
 		override, hasOverride = a.modes.ModeHotkeyOverride(key)
 	}
 
-	actions, repeat := a.effectiveHeldHotkey(hasOverride, override, globalActions, a.configSnapshot())
+	actions, repeat := a.effectiveHeldHotkey(
+		hasOverride,
+		override,
+		globalActions,
+		a.configSnapshot(),
+	)
 	if repeat {
 		a.startHotkeyRepeat(key, actions)
 
--- a/internal/app/hotkeys.go
+++ b/internal/app/hotkeys.go
@@ -71,18 +71,26 @@ func (a *App) registerHotkeys() {
 
 		var registerHotkeyErr error
 
-		if releaseManager, ok := a.hotkeyManager.(HotkeyReleaseService); ok &&
-			a.hotkeyActionsRepeatWhileHeld(bindActions, cfg) {
+		// When the backend can report key releases, register every hotkey through
+		// the release path and decide press-by-press whether to repeat. The
+		// repeat-vs-once choice depends on the effective binding (the per-mode
+		// override when the active mode binds this key, otherwise the global
+		// binding), which is only known at press time — so it cannot be made here,
+		// at registration, from the global action alone.
+		if releaseManager, ok := a.hotkeyManager.(HotkeyReleaseService); ok {
 			_, registerHotkeyErr = releaseManager.RegisterWithRelease(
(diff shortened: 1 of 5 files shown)


y3owk1n:
Thanks for the PR for fixing this, it was on my list but was abandoned due to priority. Previously our global hotkeys was implemented with carbon hotkeys while event taps hotkeys are with cgevents. At PR #967, we had migrated all of the carbons to cgevents, and this fix definitely comes in at a good timing.

I've done some manual fix ups to satisfy linting, the changes overall are quite straight forward and tests are included too, nice work. I have no comment on the code part.

Next we'll be running the CI and some greptile bug catcher, we are good to go if everything passes.

y3owk1n:
@greptileai

y3owk1n:
Everything looks good, we are merging and thanks! It will be available on nightly channel in a bit, but for stable release, it would be most likely on next monday.

gabrielecirulli:
Thanks for the quick turnaround!

PR #1000, opened 2026-07-09
chore(just): add install recipe for every platforms

gabrielecirulli:
Disclaimer: This code was written by Claude. I steered it toward a solution I thought sensible.

To be clear on the intent for this PR, I don't expect this to be merged, just wanted to share back my work in case it's useful.

## Rationale

I run Neru from source because I want to make my own changes as I use it, instead of waiting for them to land upstream. So I wanted one command that installs a source build properly, the way a package would: the app in /Applications, the login agent registered, `neru` linked onto PATH, and completions and man pages in place. The documented `just release` then `mv` steps only copy a file and skip all of that. It should be handy for anyone else hacking on Neru too.

## What it does

`just install` dispatches on the host OS and walks through the steps, asking before each one.

- macOS: copies the app bundle to /Applications, registers the launchd login agent, links `neru` onto PATH, and installs shell completions and man pages. It stages the bundle so an overwrite never wipes the old install first, stops a running daemon before replacing it, and refuses to run over a Homebrew cask or a nix-darwin/home-manager install, pointing you at the right update or uninstall command.
- Linux: copies the binary to ~/.local/bin, then offers a systemd user service (the shape the Nix modules use), the input group for Wayland, completions, and man pages.
- Windows: minimal. Copies the exe under %LOCALAPPDATA% and offers a per-user autostart Run entry.

## Docs

- Document `just install` in INSTALLATION.md, with the manual steps kept as a fallback, add it to the recipe table in DEVELOPMENT.md, and name it in the README install section.
- Fix the systemd example in LINUX_SETUP.md, which used `neru daemon` (not a real command, so it would fail-loop under Restart=on-failure) and carried a `Nice=-10` that fails an unprivileged user service.

## Notes

- I have only run the macOS branch on a real machine. The Linux and Windows branches are unverified.

## Potential improvements

- Non-macOS launch still prints "Only macOS is currently supported" (in `cmd/neru/main.go:52-54`, the `LaunchDaemon` warning) while the README and CROSS_PLATFORM tables claim Linux and Windows support. That contradiction predates this change.
- The Nix modules still set `Nice=-10` and rely on `graphical-session.target`, the same reliability caveats this recipe works around. Worth a separate pass.


gabrielecirulli:
The `install` in justfile is quite huge. I might break it into a separate script.

y3owk1n:
> The `install` in justfile is quite huge. I might break it into a separate script.

Agreed, i think we should put it in `root/scripts` directory. Maybe can split them into:

- `root/scripts/install-darwin.sh`
- `root/scripts/install-linux.sh`
- `root/script/install-windows.sh`

gabrielecirulli:
Pushed a commit that breaks the code into 3 scripts.

gabrielecirulli:
Converted to draft as I would like to add a `-y` flag to install script

gabrielecirulli:
Done. 

My only remaining question is whether the output of the just install command is acceptable. It's progressive as there are various questions asked as you go but it does feel like a wall of text.

CleanShot 2026-07-09 at 14 35 35@2x

y3owk1n:
> Done. 
> 
> 
> 
> My only remaining question is whether the output of the just install command is acceptable. It's progressive as there are various questions asked as you go but it does feel like a wall of text.
> 
> 

Personally I think it's fine. Since it touches user's system, it's better to be verbose than not. Most install script are as verbose as this.

y3owk1n:
@greptileai

gabrielecirulli:
Pushed a fix to @greptileai's comment

y3owk1n:
everything looks good to me at this point, will merge when ci done. Thanks!

Issue #998, opened 2026-07-09
Create a custom keyboard shortcut for a specific UI element


gabrielecirulli:
### Pre-flight checks

- [x] I have searched existing issues and discussions and this has not been requested yet

### Category

Navigation (hints, grid, recursive grid, scroll)

### Target platform

All platforms

### Problem or use case

Recently, Claude for macOS removed the ⌘+1, ⌘+2, ⌘+3 shortcuts to switch between Claude, Cowork, and Code.

### Proposed solution

This got me thinking whether it'd make sense to have a feature that lets you bind a keyboard shortcut to an action on a specific element on screen, targeted by either its assistive ID (if stable) or by OCR. 

This would let you define shortcuts for common actions in apps that wouldn't otherwise let you do so.

I don't know if this is already possible through a clever configuration. Curious about your thoughts on this.

### Contribution

- [ ] I'm willing to submit a PR for this feature

y3owk1n:
Hi there, i added a cool new flag as you see above (not on stable release yet, it's available in nightly now) that might be able to support what you need. Generally there are 2 ways to achieve your goal.

### 1. Absolute mouse coordination (available in stable)

If you are always on a fixed screen and never resize or reposition the window, the most direct way is to just force a click on a x and y coordinate.

```toml
"Cmd+1" = ["action move_mouse --x 113 --y 123", "action left_click"]
```

### 2. Tap into accessibility (available in nightly)

If the above is a no-go, the only way is to tap into our hints mode (you have to enable it), and do some magic around it, it is possible now after #999 implementation.

But this method relies a lot on how the whole accessibility of the program structured. This doesn't always work well in my opinion, mainly due to how common are the text, and i dont think there's any stable reference that we can target for this case. But the following snippet should work if we only have a single button after the filter, and we can feed `a` key to the mode (select the 1st label, depends on your hints configuration setup tho) 

docs about feed action -> https://github.com/y3owk1n/neru/blob/main/docs/CLI.md#feed-keys

```toml
"Cmd+1" = ["hints --role AXButton --text Home --action left_click", "action feed --mode a"]
"Cmd+2" = ["hints --role AXButton --text Code --action left_click", "action feed --mode a"]
```

https://github.com/user-attachments/assets/3c3383bc-bc59-4bdf-abe6-d542fab463cf

Even with OCR, these text are way too common and might have lots of potential duplication in the whole program UI.

[...]

PR #1004, opened 2026-07-09
docs(tips): add mode-cycling and element-targeted shortcut recipes

gabrielecirulli:
Disclaimer: This doc was written by Claude. I steered it toward the setup I actually use.

Follow-up to #998, where I offered to add a couple of recipes to the tips docs. Here they are.

## Rationale

Two things I set up for myself and figured were worth writing down: cycling every navigation mode from one key, and faking a shortcut for a UI element an app never exposes (the case that kicked off #998, when Claude for macOS dropped its Cmd+1/2/3 view switch).

## Changes

Two new sections in `docs/TIPS_TRICKS.md`:

- **Cycle Through Modes with One Hotkey** — bind a single key to walk hints → recursive grid → grid → scroll and wrap back to hints. It works because a per-mode hotkey overrides a global hotkey bound to the same key (#993): the global binding fires from idle, and each mode's own binding takes over once you are inside it.
- **Bind a Shortcut to a Specific UI Element** — rebuild a shortcut an app never gives you. The binding lives in a root-level `[[app_configs]]` block (#1008) scoped to the target app by `bundle_id`, so it drives Neru only in that app and passes through everywhere else. Inside the block it clicks the element one of three ways: absolute coordinates, window-relative coordinates, or filtered hints plus `action feed --mode`. Ends with a Claude view switcher binding Cmd+1/2/3.

Also adds the existing "Mode Toggle (On/Off)" section to the table of contents, which was missing its entry.

## Addressing the review

@y3owk1n pointed out that the element recipe's app-scoping steps read like a temporary workaround, since scoping a global hotkey to one app is something Neru should do itself. #1008 landed exactly that with root-level `[[app_configs]]`, so I dropped the three workaround sections (the external-tool scoping note, the bash view-switcher script, and the per-platform list of Karabiner / AutoHotkey / GNOME shortcut tools) and rewrote the recipe to use `[[app_configs]]` directly.

## Notes

- These recipes rely on features that are on `main` but not in a tagged release yet: #993 for the per-mode hotkey override, #999 for `action feed --mode`, and #1008 for scoping global hotkeys to one app. Each section notes it needs 1.47.0 and later. If that version number is wrong, tell me and I'll fix it.

### DIFF
--- a/docs/TIPS_TRICKS.md
+++ b/docs/TIPS_TRICKS.md
@@ -12,6 +12,9 @@
 - [Custom Mouse Movement Step Size](#custom-mouse-movement-step-size)
 - [Click, sleep, move](#click-sleep-move)
 - [Target Menus Without Moving the Real Cursor](#target-menus-without-moving-the-real-cursor)
+- [Mode Toggle (On/Off)](#mode-toggle-onoff)
+- [Cycle Through Modes with One Hotkey](#cycle-through-modes-with-one-hotkey)
+- [Bind a Shortcut to a Specific UI Element](#bind-a-shortcut-to-a-specific-ui-element)
 - [Disabling All Built-In Hotkeys](#disabling-all-built-in-hotkeys)
 - [Give Browser Content Time To Load Before Refreshing Hints](#give-browser-content-time-to-load-before-refreshing-hints)
 - [Checking the Accessibility Tree on macOS](#checking-the-accessibility-tree-on-macos)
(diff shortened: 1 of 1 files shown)


y3owk1n:
Overall i think its good, except i am thinking do we really want these 3 sections:

- `Scoping the trigger to one app`
- `Example: a Claude view switcher`
- `Hotkey tools per platform`

Rationale behind:

`Scoping the trigger to one app` is something that Neru should provide but was forgotten, and it wouldn't be hard to support this in the future. These sections feels more like a temporary workaround than tips and tricks to me. Maybe we should take them out and add a new issue for global hotkey app configs implementation.

What's your thought?

gabrielecirulli:
If that's on the roadmap, then I agree it should be removed and maybe replaced with a note that says it's not supported yet but will be possible in the future.

y3owk1n:
> If that's on the roadmap, then I agree it should be removed and maybe replaced with a note that says it's not supported yet but will be possible in the future.

Check out my recent merged PR #1008, now app based scoping is possible, most likely we dont need to have those sections, that what i meant ~

gabrielecirulli:
Done. Dropped the three workaround sections (the external-tool scoping note, the bash view-switcher script, and the per-platform hotkey tool list) and rewrote the element recipe to use the root-level `[[app_configs]]` from #1008, scoped by `bundle_id`. The three targeting approaches (absolute coords, window-relative coords, filtered hints + `action feed --mode`) now live inside `[[app_configs]]` blocks, ending with a Claude view switcher on `Cmd+1/2/3`.


One thing to check: I gated the new `[[app_configs]]` note at "1.47.0 and later" to match the mode-cycling section, since #1008 isn't in a tagged release yet. If that version is wrong, let me know and I'll fix it.

y3owk1n:
LGTM, will merge when CI done.

PR #1012, opened 2026-07-10
fix(hints): keep labels on screen during refresh instead of blanking

gabrielecirulli:
Disclaimer: This code was written by Claude. I steered it toward a fix I think is right and verified it behaves.

## What this does

Entering a hints refresh used to blank all the labels for the length of the re-scan, so they flickered out and back in. This keeps them on screen and morphs them in place, so a refresh is seamless.

## Why

On a refresh, `activateHintModeInternal` cleared the overlay before the slow re-scan (100-300ms), which is the blank gap. Removing that pre-scan clear is the whole fix. The overlay's existing incremental draw then updates the labels in place with no blank frame.

## The change

- Do not clear the overlay on the refresh path. The existing labels stay until the fresh scan updates them in place. A fresh (non-refresh) activation still clears as before, and a refresh that resolves to zero hints still clears through the normal exit path.
- The screen-capture-permission bail now exits through the normal exit path on a refresh, so it clears the stale overlay like the other refresh abort paths (the pre-scan clear used to cover it).

The net change is only in `internal/app/modes/hints.go`.

## Note on the commits

The history has a detour. An earlier commit also removed the overlay's incremental structural draw path, and a later commit restores it. That path is the live search-filter redraw (tuned in #442 and #443 as a per-keystroke hot path for many hints), so it stays. The net change of this PR is only the clear removal.

## Scope

This is independently mergeable. It is also groundwork for upcoming push-based hint auto-refresh (a `hints --watch` mode) that is not published yet, where a seamless in-place refresh matters more. That work lands as its own PRs later.

## Testing

Built and ran locally. Entered hints, triggered refreshes (modifier passthrough and manual), and the labels stay put and morph in place with no blank frame. Normal hints entry, prefix typing, and search filtering still work. `go build ./...`, `go vet`, and the `internal/app/modes` tests pass.


gabrielecirulli:
Claude got too eager with removing code. It looks like the `drawHintsIncrementalStructural` was useful for the hot path of typing to search. I've made it restore that. I'm unsure whether this optimization actually yields significant performance improvements compared to fully redrawing but might not be worth the risk of removing it in this PR.

y3owk1n:
This is smooth, overall scoped and straight forward. We'll merge after CI and greptile passes.

y3owk1n:
@greptileai

PR #1016, opened 2026-07-12
refactor(config): rename hints.additional_ax_support to hints.web_content_hints

gabrielecirulli:

## Rationale

The `[hints.additional_ax_support]` section only controls whether hints reach web-page content inside Chromium and Firefox browsers. Now that Electron apps are woken without any setting (#1015), the old name no longer describes what the section does.

## What this changes

Rename the section to `[hints.web_content_hints]`, which is self-describing, and rename its on/off field from `enable` to `enabled` to match the boolean naming used elsewhere in the config (`grid.enabled`, `systray.enabled`). The four bundle lists keep their names.

neru's TOML loader ignores unknown keys, so a config still using the old section or the old field name would silently lose its settings. To prevent that, the loader rejects both through the existing configuration-validation error path, the same one that shows the "Configuration Validation Failed" dialog, with a message naming the new key.

## Changes

- `config.go`: rename the `AdditionalAXSupport` type to `WebContentHints` and its `Enable` field to `Enabled`, updating the toml/json tags to `web_content_hints`/`enabled`.
- `config_defaults.go`, `lifecycle.go`, `tree.go`: update every reader.
- `service.go`: reject both the removed section and the removed `enable` field during load with a `CodeInvalidConfig` error that points at the new key.
- `service_test.go`: cover the rejection of the old section and the old field, and a successful load of the new names.
- Configs and docs: rename the section and field everywhere they appear.

## Breaking change

The `[hints.additional_ax_support]` section is renamed to `[hints.web_content_hints]` and its `enable` field to `enabled`. A config that still uses either old name fails to load with a validation error telling the user to rename it. The four bundle lists keep their names.

## Testing this change

The rename and its load-time checks can be verified on this branch alone (build `refactor/rename-web-content-hints`); they do not need the rest of the stack. `neru config validate` checks a config file without starting the daemon, and `neru config reload` applies it live.

**New names load.** Put the renamed section and field in your config:

```toml
[hints.web_content_hints]
enabled = true
```

`neru config validate` prints `Configuration is valid`. `neru config reload` (or a restart) applies it. With `enabled = true` neru sets `AXEnhancedUserInterface` on Chromium and Firefox so hints reach web-page content; with `enabled = false` they do not.

**Old section name is rejected.** Use the pre-rename section:

```toml
[hints.additional_ax_support]
enabled = true
```

`neru config validate` fails, and a live `neru config reload` rejects it the same way and keeps the previous/default config rather than silently dropping the setting:

```
Configuration validation failed:

  [hints.additional_ax_support] has been renamed to [hints.web_content_hints]. Rename that section in your config to keep browser web-content hints.
```

**Old field name is rejected.** Use the new section but the old `enable` field:

```toml
[hints.web_content_hints]
enable = true
```

Same rejection, pointing at the field:

```
[hints.web_content_hints] enable has been renamed to enabled. Rename that key in your config to keep browser web-content hints.
```

## Testing the whole feature

This PR is a prerequisite and does not exercise the feature on its own. To test auto-refresh end to end, pull and run the tip branch `feat/hints-auto-refresh-config` (#1018), which stacks all four changes. Its description lists the exact config to set and the steps to follow.

## Merge order

These four PRs are one series and must merge in order, each only after the one before it lands:

1. #1015 — `refactor/ax-enablement-manual-on-all` — show hints in any Electron app without whitelisting it
2. #1016 — `refactor/rename-web-content-hints` — rename `hints.additional_ax_support` to `hints.web_content_hints`  ← this PR
3. #1017 — `feat/ax-observer-service` — push-based AX observer service (inert)
4. #1018 — `feat/hints-auto-refresh-config` — opt-in hints auto-refresh

Each targets `main`, so until the ones before it merge, its diff also shows their commits.

### DIFF
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -101,7 +101,7 @@ visible_check_enabled = false
 "Right" = "action move_mouse_relative --dx=10 --dy=0"
 
 [hints.additional_ax_support]
-enable = false                              # Enable enhanced AX for Electron/Chromium/Firefox/WebKit apps
+enable = false                              # Chromium/Firefox web-page hints via AXEnhancedUserInterface. Electron works without it
 additional_electron_bundles = []
 additional_chromium_bundles = []
 additional_firefox_bundles = []
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -100,8 +100,8 @@ visible_check_enabled = false
 "Left" = "action move_mouse_relative --dx=-10 --dy=0"
 "Right" = "action move_mouse_relative --dx=10 --dy=0"
 
-[hints.additional_ax_support]
-enable = false                              # Chromium/Firefox web-page hints via AXEnhancedUserInterface. Electron works without it
+[hints.web_content_hints]
+enabled = false                             # Chromium/Firefox web-page hints via AXEnhancedUserInterface. Electron works without it
 additional_electron_bundles = []
 additional_chromium_bundles = []
 additional_firefox_bundles = []
--- a/configs/hints-only-config.toml
+++ b/configs/hints-only-config.toml
@@ -51,8 +51,8 @@ ignore_clickable_check = false
 # 	"Shift+L" = "__disabled__"
 # }
 
-[hints.additional_ax_support]
-enable = true
+[hints.web_content_hints]
+enabled = true
 
 additional_electron_bundles = []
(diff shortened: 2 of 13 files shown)


gabrielecirulli:
@greptileai ?

gabrielecirulli:
Superseded by #1035

PR #1017, opened 2026-07-12
feat(axobserver): add push-based AX observer service

gabrielecirulli:
Disclaimer: This code and this PR description were partly written by AI.

## Rationale

This is the engine for hint auto-refresh. It's not actually used in Neru's hint mode yet. It's split out so the feature PR (#1018) stays focused on the hints part of the implementation.

## What this adds

macOS allows you to register observers to tell when an app's UI changes. This PR adds the observer setup but does not connect it to hints mode yet. That's done in #1018.

The notifications we want to get are selected via a bitmask. We select most notification types, including ones that browsers post for web-page content.

Notifications are delivered to a background thread waiting for OS events. The Go side owns the thread. Adding or removing a watcher is a command to the thread. When changes fire, the thread sends a callback to Go with the pid and the name of the change.

The rest of neru accesses this through the `Manager` object. You pass the set of apps you want watched and call `Reconcile`. It compares that set against the current watchers. It sets up new ones or tears down ones no longer in use.

Until `Reconcile` is called with at least one target, nothing is armed and no OS resources are held, which is why this PR is inert on its own.

## Other platforms

On any non-macOS platform, the `Manager` is a stub that does nothing, so the rest of neru still compiles and runs. 

## Tests

Cover the Manager against a fake platform. Adds a goleak guard on the package; and a darwin soak that arms and disarms 1000 times, asserting zero live Core Foundation objects and no run-loop thread at idle after each cycle.

## Testing the whole feature

The actual feature is implemented in #1018. This PR is a prerequisite.

## Merge order

This PR and the feature PR are a two-part stack and must merge in order:

1. #1017 — `feat/ax-observer-service` — push-based AX observer service (inert)  ← this PR
2. #1018 — `feat/hints-auto-refresh-config` — opt-in hints auto-refresh

Both target `main`, so until this one merges, #1018's diff also shows this PR's commits.

Two earlier PRs (#1015 and #1016) have been closed. #1035 removed the `additional_ax_support` / `web_content_hints` machinery they depended on, so both became unnecessary, and this branch was rebased directly onto `main` without them.


gabrielecirulli:
@greptileai ?

internal/core/infra/axobserver/manager.go
```diff
@@ -0,0 +1,140 @@
...
+	Mask Mask
+}
+
+// Manager keeps the set of armed observers reconciled with the desired targets.
+//
+// It is single-owner: Reconcile, DisarmAll, and Close must be called from one
+// goroutine, or otherwise serialized by the caller. The change callback passed
+// to New is invoked from the platform's callback thread; it must be cheap and
+// must not call back into the Manager (that would deadlock against a concurrent
+// teardown).
```
gabrielecirulli: This comment is unintelligible. Simplify and make more concise.

internal/core/infra/axobserver/manager.go
```diff
@@ -0,0 +1,140 @@
...
+	platform Platform
+	logger   *zap.Logger
+
+	mu     sync.Mutex
+	armed  map[int]Mask
+	closed bool
+}
+
+// New creates a Manager for the current platform. onChange receives the pid of
+// any application whose UI structurally changes while its observer is armed.
```
gabrielecirulli: Explain that `onChange` is passed by the user to receive callbacks with pids of applications whose ui changes while observer is armed

gabrielecirulli:
@greptileai

internal/core/infra/axobserver/manager.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	desired := make(map[int]Mask, len(targets))
+	for _, target := range targets {
+		if target.PID <= 0 || target.Mask == 0 {
+			continue
+		}
+
+		desired[target.PID] = target.Mask
+	}
```
gabrielecirulli: Can this be written more like a JavaScript array filter call kind of thing?

internal/core/infra/axobserver/platform.go
```diff
@@ -0,0 +1,28 @@
...
+	// Disarm removes the observer for pid, if any.
+	Disarm(pid int)
+
+	// DisarmAll removes every observer and releases any run-loop resources.
+	DisarmAll()
+
+	// SetSink registers the change callback, which receives the firing pid and
+	// the name of the notification that fired (for debug logging). Passing nil
+	// clears it so no further callbacks are delivered.
+	SetSink(func(pid int, notif string))
```
gabrielecirulli: Can this not have a less cryptic name?

internal/core/infra/axobserver/platform_darwin.go
```diff
@@ -0,0 +1,157 @@
...
+		{notifMenuClosed, darwin.AXNotifMenuClosed},
+		{notifFocusedUIChanged, darwin.AXNotifFocusedUIChanged},
+		{notifValueChanged, darwin.AXNotifValueChanged},
+		{notifLiveRegionChanged, darwin.AXNotifLiveRegionChanged},
+		{notifLiveRegionCreated, darwin.AXNotifLiveRegionCreated},
+		{notifExpandedChanged, darwin.AXNotifExpandedChanged},
+		{notifRowExpanded, darwin.AXNotifRowExpanded},
+		{notifRowCollapsed, darwin.AXNotifRowCollapsed},
+		{notifElementBusyChanged, darwin.AXNotifElementBusyChanged},
+	}
```
gabrielecirulli: I don't think we need to store masks in detail on the go side and have machinery to translate them to the native side. The go side should not even care about masks at all. The darwin side beyond the bridge should be the one defining the masks, and there should only be one mask in the end. We don't need to have such a complicated setup.

Maybe we don't even need a mask, because we're actually referring to ax elements on the darwin side. We probably need to remove the entire concept of a mask all the way upstream into the manager, because we built infrastructure to track masks, but that's not even an option that we should surface.

internal/core/infra/axobserver/manager.go
```diff
@@ -0,0 +1,140 @@
...
+import (
+	"sync"
+
+	"go.uber.org/zap"
+)
+
+// Target names a pid to observe and the structural notifications to watch for.
+type Target struct {
+	PID  int
+	Mask Mask
```
gabrielecirulli: Mask should not be an exposed option from the Manager. You track a PID, you get changes. How the changes are generated or how the mask is defined is only a concern of the bottom layer that interfaces with the operating system itself, and it is not configurable.

internal/core/infra/axobserver/masks.go
```diff
@@ -0,0 +1,46 @@
...
+	notifMenuOpened
+	notifMenuClosed
+	notifFocusedUIChanged
+	notifValueChanged
+	notifLiveRegionChanged
+	notifLiveRegionCreated
+	notifExpandedChanged
+	notifRowExpanded
+	notifRowCollapsed
+	notifElementBusyChanged
```
gabrielecirulli: Remove mask definitions from the Go side.

internal/core/infra/platform/darwin/axobserver.go
```diff
@@ -0,0 +1,112 @@
...
+	AXNotifMenuClosed         AXObserverMask = C.NERU_AXNOTIF_MENU_CLOSED
+	AXNotifFocusedUIChanged   AXObserverMask = C.NERU_AXNOTIF_FOCUSED_UI_CHANGED
+	AXNotifValueChanged       AXObserverMask = C.NERU_AXNOTIF_VALUE_CHANGED
+	AXNotifLiveRegionChanged  AXObserverMask = C.NERU_AXNOTIF_LIVE_REGION_CHANGED
+	AXNotifLiveRegionCreated  AXObserverMask = C.NERU_AXNOTIF_LIVE_REGION_CREATED
+	AXNotifExpandedChanged    AXObserverMask = C.NERU_AXNOTIF_EXPANDED_CHANGED
+	AXNotifRowExpanded        AXObserverMask = C.NERU_AXNOTIF_ROW_EXPANDED
+	AXNotifRowCollapsed       AXObserverMask = C.NERU_AXNOTIF_ROW_COLLAPSED
+	AXNotifElementBusyChanged AXObserverMask = C.NERU_AXNOTIF_ELEMENT_BUSY_CHANGED
+)
```
gabrielecirulli: Remove the mask set from this side of the bridge.

internal/core/infra/platform/darwin/axobserver_darwin.m
```diff
@@ -0,0 +1,337 @@
...
+	table[9] = (NeruNotifEntry){NERU_AXNOTIF_FOCUSED_UI_CHANGED, kAXFocusedUIElementChangedNotification};
+	table[10] = (NeruNotifEntry){NERU_AXNOTIF_VALUE_CHANGED, kAXValueChangedNotification};
+	// Web-content signals. Chromium and Firefox post these string notifications
+	// (no public kAX* constant for the first three) when a page's content changes.
+	table[11] = (NeruNotifEntry){NERU_AXNOTIF_LIVE_REGION_CHANGED, CFSTR("AXLiveRegionChanged")};
+	table[12] = (NeruNotifEntry){NERU_AXNOTIF_LIVE_REGION_CREATED, CFSTR("AXLiveRegionCreated")};
+	table[13] = (NeruNotifEntry){NERU_AXNOTIF_EXPANDED_CHANGED, CFSTR("AXExpandedChanged")};
+	table[14] = (NeruNotifEntry){NERU_AXNOTIF_ROW_EXPANDED, kAXRowExpandedNotification};
+	table[15] = (NeruNotifEntry){NERU_AXNOTIF_ROW_COLLAPSED, kAXRowCollapsedNotification};
+	table[16] = (NeruNotifEntry){NERU_AXNOTIF_ELEMENT_BUSY_CHANGED, kAXElementBusyChangedNotification};
```
gabrielecirulli: Since we don't want to allow a mask as an input to the whole system anymore, let's maybe get rid of this table and pass the notification types directly.

gabrielecirulli (commented):
I am worried that the shape of allowing to pass a PID and tracking independent PIDs for notifications doesn't really match what we're going to be doing downstream in PR #1018. We're probably only going to be tracking one PID at a time, depending on what's focused, so all of this architecture maybe doesn't make a lot of sense. 

EDIT: Confirmed in #1018 at updateAutoRefreshObservers: it's only called in one place and with one pid. The whole reconciliation and multi-pid machinery is useless. Reorganize and refocus the code into a one-pid-at-a-time structure

gabrielecirulli:
@greptileai

PR #1018, opened 2026-07-12
feat(hints): opt-in auto-refresh that keeps hints live on macOS

gabrielecirulli:

## Rationale

Hints are scanned once when hints mode opens, so they go stale when the focused window changes afterward: a page finishes loading, a menu opens, a panel appears. This wires the AX observer service (#1017) to a new opt-in setting that re-scans hints when the focused app's UI changes, with no keypress. macOS only for now; on other platforms the observer backend is a stub and the setting is inert.

## What this adds

**The setting.** `[hints.auto_refresh]` has two keys: `enabled` (off by default) and `debounce_ms` (how long a burst of changes settles before the re-scan, default 150). The set of accessibility notifications that wake a re-scan is not configurable: the observer watches a fixed `DefaultMask` defined in the `axobserver` package (#1017), which covers the structural notifications plus the web-content signals browsers post when a page changes, and omits `value_changed` because it fires continuously.

**Debounce and coalescing.** The refresh paths that re-enter through `activateHintModeInternal` (a `--repeat` re-entry, a modifier-passthrough re-scan, a bound `hints` re-launch, a cycle-hint re-scan, and the debounced observer) coalesce there, as Handler state rather than a separate coordinator:

- Leading edge: the first refresh in an idle burst scans immediately, so a manual refresh is never delayed.
- Trailing edge: refreshes arriving during the debounce window collapse into a single follow-up scan.
- Max-wait cap: a continuously-changing app still refreshes at a bounded rate.

While a search query is being typed, a refresh is held and released when the search ends, so the query and its filtered hint set are not swapped out from under the keystrokes. A pending hint-label selection is not held, because the `--repeat` re-scan right after a label is chosen must proceed or the overlay freezes on the stale filter.

**Backoff until the page settles.** An observer-driven scan can catch a web page mid-render and label only part of it. Browsers post a notification when content starts changing but none when it finishes, and single-page sites load content in waves a second or more after the click. So after each observer-driven scan, neru keeps re-scanning on a capped exponential backoff. Each re-scan fingerprints the resulting hint set (a hash of element bounds and roles, from hints the scan already produced, so it costs no extra accessibility calls). If the set changed, the interval restarts dense (250ms); if it is stable, the interval grows by 1.6x up to a 5s ceiling; the loop stops after two consecutive stable scans at that ceiling. Two ceilings bound a page that never settles: the scan count (25) and the elapsed window (30s) since the sequence began do not reset on change, so a continuously-changing live page winds down to dormant rather than re-scanning forever, until the next observer event starts a fresh sequence.

A manual or `--repeat` refresh takes priority over a running settle loop: it ends the settle and scans immediately as the leading edge, so selecting a hint while the settle is mid-flight is never deferred behind a pending re-check.

**Safety.** The observer callback runs on the accessibility thread and touches only a leaf mutex, never the mode lock, so it cannot deadlock a teardown that joins that thread. Observers arm on the focused app while hints are open and are disarmed on exit, so an idle neru holds no accessibility resources.

## Tests

Cover the debounce (leading, trailing, max-wait, hold-during-search, no-hold-for-label), the backoff interval growth and clamp, the stop conditions (two stable at the cap, the scan-count ceiling, the window ceiling), the state machine, and a manual refresh pre-empting the settle.

## Notes

This sits on top of the merged label-persistence fix (#1012). A stable re-scan that changes nothing redraws through the same incremental path, so it is not visible.

## Testing the whole feature

This is the tip of the two-part stack (#1017 then this) and sits on top of `main`, so it is the branch to run to test the feature end to end. Pull and run `feat/hints-auto-refresh-config` (#1018).

macOS only. Set this in your neru config:

```toml
[hints.auto_refresh]
enabled = true          # re-scan hints when the focused app's UI changes
```

`hints.auto_refresh.enabled = true` is the only switch the feature needs. It refreshes native windows, menus, and browser and Electron web-page content alike: neru wakes the focused app's accessibility tree on focus, so page-content changes reach the observer without any extra setting. `debounce_ms` can stay at its default.

Then open hints mode and watch the labels update as the focused app changes, a page finishes loading, a menu opens, or a panel appears, with no keypress.

## Potential improvements

- The fingerprint hashes exact element bounds, so one continuously-animating element on a page keeps the backoff pinned dense until the 25-scan ceiling instead of settling. Quantizing bounds to a coarse grid before hashing would let such a page settle early.
- On multi-monitor setups the re-scan resolves the target screen from the cursor location, the same as a fresh activation. Moving the cursor to another display during a settle can filter the window's hints off screen for that scan. Pinning a refresh to the screen the session started on would fix it.

## Merge order

This PR and the observer PR are a two-part stack and must merge in order:

1. #1017 — `feat/ax-observer-service` — push-based AX observer service (inert)
2. #1018 — `feat/hints-auto-refresh-config` — opt-in hints auto-refresh  ← this PR

Both target `main`, so until #1017 merges, this PR's diff also shows its commits.

Two earlier PRs from the original series have been closed: #1015 (show hints in any Electron app without whitelisting) and #1016 (rename `hints.additional_ax_support` to `hints.web_content_hints`). Upstream's bundle-type auto-detection (#1035) removed the `additional_ax_support` / `web_content_hints` machinery they depended on, so both became unnecessary; this branch was rebased directly onto `main` without them, and the feature no longer needs any browser-specific setting to reach web-page content.


gabrielecirulli:
@greptileai ?

Issue #1019, opened 2026-07-12
Missing hints on some web pages


gabrielecirulli:
### Pre-flight checks

- [x] I have searched existing issues and this bug has not been reported yet

### Neru version

`main` (`c28300bc24773815f1c37d22ee4d70023247650e`)

### Operating system

macOS

### OS version

macOS 26.5.1

### Navigation mode

Hints

### What happened?

On some websites like news.ycombinator.com, hints are missing for some of the links, for example, the comments link in the screenshot. I was not able to find a setting to change this. 

<img width="3644" height="2370" alt="Image" src="https://github.com/user-attachments/assets/5554c539-f74e-417a-8269-bf98d69bd6e7" />

### What did you expect?

Hints should appear for every available link or button. 

### Steps to reproduce

Run Hints mode on a website such as news.ycombinator.com. 

### Config (relevant sections)

```toml

```

y3owk1n:
I think this is very much from the heuristic check here https://github.com/y3owk1n/neru/blob/main/internal/core/infra/accessibility/tree.go#L701C1-L702C1.

If you try to open the same link using safari browser, you should be seeing all of the hints. We had included a heuristic check of 15px to be excluded, because electron or chromium has a very noisy ax tree, without early return, it will go through the whole noisy tree, and it's unbearable slow... this is not an issue for safari, as safari tree is generally pruned and nicer to work with. This is also part of all the optimisation that I had implemented to make the hints mode as fast as possible.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/93711d51-fd48-4a30-87e2-371296c6bb3c" />

If you look at `AXSize`, you'll see the height is `11`, which is lower than our heuristic set at `15`.

You can try to bump the `minElementSize` to `10`, and it should work fine on chromiums.

### Solution

- I could either manually lower this to `10` as hardcoded
- I could expose this to a config field within `additional_ax_support`, with a default of `15` (so no breaking for existing users), and user can decide to bump it lower, but noisier.

y3owk1n:
i think manually lowering down to `10` might be the best fix for this, as hackernews small link is approximately 9px after computed, which is probably a common lowest font size that normal browsers will clamp to (as far as I remember).

try applying this patch and see if it works:

```diff
diff --git c/internal/core/infra/accessibility/tree.go w/internal/core/infra/accessibility/tree.go
index 71e7280d..6aabb8e4 100644
[...]

PR #1029, opened 2026-07-13
docs(troubleshooting): Add notes about element size and element type, clean up Electron/Browser troubleshooting

gabrielecirulli:
Disclaimer: This PR description was written by Claude. I reviewed it and it matches the change.

## Context

Follow-up to #1019, where hints were missing for some links on certain web pages (for example the small comment links on Hacker News). The root cause was fixed in code by lowering the minimum element size threshold, and the phantom hint clusters that appeared afterward were traced to the `AXRow`/`AXCell` clickable roles being applied to table-based layouts. This adds the troubleshooting entry the maintainer asked for so users hitting the same symptoms can self-serve.

## What Changed

All edits are in `docs/TROUBLESHOOTING.md`:

- Add "Some elements that should have hints don't have hints", covering the two causes: the element is too small to pass the size filter (open an issue), or a relevant role is not in `hints.clickable_roles` (links to `default-config.toml` for the full list of roles).
- Add "Certain hints don't visually match any on-screen UI", explaining that Neru may generate hints for non-visible structural elements such as `AXRow` and `AXCell`, and that removing those roles and running `neru config reload` clears them.
- Consolidate the Electron and browser hint troubleshooting. The same topic was previously split across three headings ("Electron/Chromium/Firefox issues", "Hints don't appear in Electron apps", "Hints don't appear in Chrome/Firefox content"), two of which were nearly empty. These are merged into a single "Hints don't appear in Electron apps or Chrome/Firefox content" section with the full config, restart, and log-check steps.
- Fix the markdown indentation of the keyboard-layout code blocks so they nest correctly under their list item.

## Notes

Documentation only. No code changes.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -152,13 +152,6 @@ Restart Neru:
 pkill neru && neru launch
 ```
 
-**Check logs for:**
-
-```
-App requires Electron support
-Enabled AXManualAccessibility for: com.your.app
-```
-
 ### Some elements that should have hints don't have hints
 
 This can happen when the element you're trying to select is small. If you encounter this, open an issue.
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -163,22 +163,29 @@ Enabled AXManualAccessibility for: com.your.app
 
 This can happen when the element you're trying to select is small. If you encounter this, open an issue.
 
-Alternatively, make sure all relevant roles are enabled. Look at [default-config.toml](https://github.com/y3owk1n/neru/blob/3a8d3c848d63e6757b602cb4dc31fb532aa310d7/configs/default-config.toml#L65) for available roles.
+Alternatively, make sure all relevant roles are enabled. Copy the complete `clickable_roles` list from [default-config.toml](https://github.com/y3owk1n/neru/blob/3a8d3c848d63e6757b602cb4dc31fb532aa310d7/configs/default-config.toml#L65) to enable all roles.
+
+Run `neru config reload` and then test.
 
 ```toml
 [hints]
 clickable_roles = [
-    "AXButton",
-	"AXMenuButton",
     # ...
 ]
 ```
 
 ### Certain hints don't visually match any on-screen UI
 
-Neru may be generating hints for elements that are not directly visible to you, such as `AXRow` and `AXCell`. Try removing those from `hints.clickable_roles`.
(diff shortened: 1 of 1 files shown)


gabrielecirulli:
@greptileai

gabrielecirulli:
@greptileai

docs/TROUBLESHOOTING.md
```diff
@@ -122,13 +122,62 @@ neru hints               # CLI works?
...
+
+```bash
+pkill neru && neru launch
+```
+
+### Some elements that should have hints don't have hints
+
+This can happen when the element you're trying to select is small. If you encounter this, open an issue.
+
+Alternatively, make sure all relevant roles are enabled. Copy the complete `clickable_roles` list from [default-config.toml](https://github.com/y3owk1n/neru/blob/3a8d3c848d63e6757b602cb4dc31fb532aa310d7/configs/default-config.toml#L65) to enable all roles.
```
y3owk1n: I think we probably don't need this line here, because it's the default if you dont declare it at all. You don't have to copy the whole block unless you need to change something like remove or add more of the roles.

docs/TROUBLESHOOTING.md
```diff
@@ -122,13 +122,62 @@ neru hints               # CLI works?
...
+```toml
+[hints]
+clickable_roles = [
+    # ...
+]
+```
+
+### Certain hints don't visually match any on-screen UI
+
+Neru may be generating hints for elements that are not directly visible to you, such as `AXRow` and `AXCell`. Copy the complete `clickable_roles` list from [default-config.toml](https://github.com/y3owk1n/neru/blob/3a8d3c848d63e6757b602cb4dc31fb532aa310d7/configs/default-config.toml#L65), then remove only those roles so the remaining defaults stay enabled.
```
y3owk1n: lets just link to the default config without line and on main, so that the link wont be a stale link.

`https://github.com/y3owk1n/neru/blob/main/configs/default-config.toml`

y3owk1n (changes requested):
Other than that, i think the rest looks good.

docs/TROUBLESHOOTING.md
gabrielecirulli: Updated to

> Alternatively, make sure all relevant roles are enabled. If you've customized `hints.clickable_roles`, you can remove that customization or restore it to the original value from [default-config.toml](https://github.com/y3owk1n/neru/blob/3a8d3c848d63e6757b602cb4dc31fb532aa310d7/configs/default-config.toml#L65).

docs/TROUBLESHOOTING.md
gabrielecirulli: Updated to

>Neru may be generating hints for elements that are not directly visible to you, such as `AXRow` and `AXCell`. Copy the complete `clickable_roles` list from [default-config.toml](https://github.com/y3owk1n/neru/blob/main/configs/default-config.toml), then remove only those roles so the remaining defaults stay enabled.

gabrielecirulli:
@y3owk1n Updated the copy. Not sure if it fully captures what you had in mind. Feel free to suggest more changes.

y3owk1n:
LGTM, we'll merge after CI checkmark.

PR #1052, opened 2026-07-18
Update README with neru installation steps

tdegrunt:
Added installation instructions for neru and zsh completion.

## Description

Some additional installation is needed


y3owk1n:
In our installation docs, we have a section for shell completion. For our readme, should we instead add another line where follow these docs to install shell completion and point them to the right section? 🤔 What do you think?

Also i think the shell completion in our installation docs are probably not so correct, maybe we can update them or at least the zsh part to match your solution, and i can update the rest of the shell completion that neru supports later on as another follow up.

PR #1055, opened 2026-07-18
fix(linux): use native relative pointer motion

### LINKED ISSUE #921: Actual pointer doesn't sync with virtual pointer
On fedora wayland niri (distro that i am using to test), i noticed that the actual pointer doesn't sync with the virtual pointer. Need research or help. 

Repro:

1. Move cursor/pointer to some other location than center
2. Activate scroll mode for example
3. And then perform `move_mouse_relative` action

Expected: pointer should move relatively from the pointer position at `step 1`

Reality: pointer move relatively from the center of the screen

floze-the-genius:
## Description

- route `move_mouse_relative` through an optional native relative-movement path
- emit `zwlr_virtual_pointer_v1.motion` directly on wlroots compositors
- preserve the existing absolute-position fallback for X11 and other platforms
- add a regression test proving native relative movement bypasses the stale cursor-position lookup

Wayland does not expose a reliable global pointer position. When Neru falls back to its screen-center cursor cache, converting a relative action into an absolute move makes the real pointer jump relative to the center. Sending a compositor-native relative event keeps movement anchored to the actual pointer position.

## Related Issues

Fixes #921

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` - New feature
- [x] `fix` - Bug fix
- [ ] `refactor` - Code restructuring (no behavior change)
- [ ] `perf` - Performance improvement
- [ ] `docs` - Documentation only
- [x] `test` - Adding or updating tests
- [ ] `chore` - Build, CI, dependencies, tooling

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags
- [x] No darwin imports from untagged shared code
- [x] Stub implementation added for non-CGO Linux builds
- [ ] N/A - This PR does not touch platform-specific code

## General Checklist

- [x] Code formatted
- [x] Linters pass
- [x] Tests pass
- [x] Build succeeds
- [x] Tests added for changed behavior
- [x] Documentation updated (not applicable)
- [x] Commit message follows Conventional Commits

## Validation

Run in a Linux CGO container with the same native packages used by CI:

- `go test ./...`
- `go test -tags=integration ./...`
- `go test -race ./...`
- `go test -tags=integration -race ./...`
- `go vet ./...`
- `golangci-lint v2.12.2 run`
- `CGO_ENABLED=1 go build ./cmd/neru`
- `clang-format --dry-run -Werror` on changed C/header files

All passed.

## Disclosure

Implementation and regression test were generated with OpenAI Codex. The change was validated against the full Linux test, race, vet, lint, formatting, and build gates listed above.

### DIFF
--- a/internal/app/services/action_service.go
+++ b/internal/app/services/action_service.go
@@ -25,6 +25,10 @@ type ActionService struct {
 	logger   *zap.Logger
 }
 
+type relativeCursorMover interface {
+	MoveCursorBy(ctx context.Context, delta image.Point) (bool, error)
+}
+
 // NewActionService creates a new action service.
 func NewActionService(
 	accessibility ports.AccessibilityPort,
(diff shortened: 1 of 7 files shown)


y3owk1n:
Verified and tested, works as expected, and the code are straightforward. Will run the CI and greptile, we are good to merge if everything is green.

Observations:

- Sticky modifiers and mode indicators are not synced with the new implemented cursor movements
- When manually moves the mouse before getting into mode, the above indicators are still not relative to the resolved cursor position.

These can be another follow up, just noting down here, so that I can remember them.

y3owk1n:
@greptileai

y3owk1n:
Good to go, thanks for the PR!

y3owk1n (approved):
looks good

y3owk1n:
I've added a follow up issue here, if you have any clue... https://github.com/y3owk1n/neru/issues/1057

PR #1070, opened 2026-07-20
feat(overlay): add configurable text outline and strokes

### LINKED ISSUE #1067: Add text outline/stroke option for overlay labels
### Pre-flight checks

- [x] I have searched existing issues and discussions and this has not been requested yet

### Category

UI / Overlays

### Target platform

All platforms

### Problem or use case

First off — thank you for Neru. The recursive grid has become how I navigate my machine, and the layered-depth config is a lovely design.

One small readability snag: overlay label text sits directly on whatever content is underneath it, so any single `text_color` will fail somewhere. Light text washes out on bright UI, dark text disappears on dark UI, and both become unreadable over photos, video, and image-heavy canvases — exactly the cases where recursive grid is most useful, since there are no AX elements to hint instead.

`label_background` does fix the contrast, and I appreciate that it's there. But the chips are visually heavy: they occlude the content I'm trying to aim at, which works a little against the point of a subdivision grid. I found myself turning them back off and living with the low contrast.

### Proposed solution

Would you consider a thin contrasting outline (stroke) around the glyphs themselves? That's the approach subtitle renderers and most screen overlays use, and it gives the same readability with far less occlusion than a filled chip.

Config sketch, following the existing naming:

```toml
[recursive_grid.ui]
text_outline_color = "#FF000000"
text_outline_width = 2
```

Ideally it would apply to `text_color` and `sub_key_preview_text_color`, and to the equivalent keys in `hints`, `grid`, `mode_indicator`, and `monitor_select` — but even `recursive_grid` alone would be a real improvement.

Fully understand if this doesn't fit what you have in mind for the overlay rendering, or if it's awkward given how the labels are drawn. No worries either way.

### Alternatives considered

- **`label_background = true`** — works, but occludes the target area as described above.
- **Per-appearance colors** (`{ light = "...", dark = "..." }`) — helps with app chrome following the system theme, but does nothing over images or video, where the background isn't tied to appearance at all.
- **Tuning `text_color` per app via `app_configs`** — too fiddly to maintain, and still breaks within a single app whose content varies.

### Environment

Neru 1.47.0, macOS 15.

### Contribution

y3owk1n:
## Description

This PR adds `text_outline_color` and `text_outline_width` options for overlay text labels.

Defaults are `text_outline_color = ""` and `text_outline_width = 0`.

> [!NOTE]
> linux and windows code are written by AI................... I hope it works..

## Related Issues

Closes #1067

## Target Platform

- [x] macOS
- [x] Linux
- [x] Windows

## Type of Change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

## General Checklist

## Screenshots / Recordings

N/A

### Rendering approach per platform

- **macOS**: `NSStrokeColorAttributeName` / `NSStrokeWidthAttributeName` on `NSMutableAttributedString` for `NSView`-based text. For `CATextLayer` (monitor select), an `NSAttributedString` with the same attributes is set as the layer's `string`.
- **Linux (X11/Wayland)**: `cairo_text_path()` + `cairo_stroke_preserve()` + `cairo_fill()` — stroke then fill within the same path.
- **Windows**: GDI `DrawTextW` into a DIB, followed by pixel-level coverage dilation for the outline pass.

### DIFF
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -108,6 +108,8 @@ padding_x = -1                              # -1 = auto based on font size
 padding_y = -1                              # -1 = auto based on font size
 border_width = 1
 placement = "bottom"                        # top, center, bottom
+text_outline_color = ""                     # Empty = no outline
+text_outline_width = 0                      # 0 = no outline (positive = stroke width)
 
 [hints.search_input_ui]
 font_size = 10
@@ -177,6 +179,8 @@ enable_gc = false                           # Periodic memory cleanup (adds CPU 
 font_size = 10
 font_family = ""
 border_width = 1
+text_outline_color = ""
+text_outline_width = 0
 
 # Recursive grid mode (recommended)
 # See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#recursive_grid
@@ -223,6 +227,10 @@ sub_key_preview = false                     # Draw miniature key grid inside eac
 sub_key_preview_font_size = 8
 sub_key_preview_autohide_multiplier = 1.5
 sub_key_preview_label_char = ""             # Override sub-key preview labels with a single character (e.g. "·"); empty = use key
+text_outline_color = ""                     # Empty = no outline
+text_outline_width = 0                      # 0 = no outline (positive = stroke width)
+sub_key_preview_text_outline_color = ""     # Sub-key preview text outline color
+sub_key_preview_text_outline_width = 0      # Sub-key preview text outline width
 
 # Virtual pointer
 # See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#virtual_pointer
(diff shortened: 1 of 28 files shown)


Franchise712:
Thank you for making the changes, and for putting the implementation together so quickly on the back of my request.

I built #1070 from source and tried it out, and you were right: it doesn't look good. I tested it on my recursive grid with red labels at `font_size = 14`, across a range of outline widths and with the label background both on and off. Thin strokes were too subtle to notice, thicker ones started eating into the letters, and I couldn't find a setting that read better than no outline at all. At that text size the outline just doesn't earn its place.

So your instinct in the issue was the right call. Sorry for sending you down this path. Happy for you to close it, and thanks again for taking the request and building it out so we could find out.

y3owk1n:
> Thank you for making the changes, and for putting the implementation together so quickly on the back of my request.
> 
> I built #1070 from source and tried it out, and you were right: it doesn't look good. I tested it on my recursive grid with red labels at `font_size = 14`, across a range of outline widths and with the label background both on and off. Thin strokes were too subtle to notice, thicker ones started eating into the letters, and I couldn't find a setting that read better than no outline at all. At that text size the outline just doesn't earn its place.

@Franchise712 Ya, that's what i figured out too when I am implementing it, but I still want to get your confirmation if it make sense, since I am primarily using grid without labels, so it didn't bother me. For us that are in community, most of us are using either empty labels or dot based labels, if you're interested to get more opinions, feel free to join our discord channel too!

> So your instinct in the issue was the right call. Sorry for sending you down this path. Happy for you to close it, and thanks again for taking the request and building it out so we could find out.

Don't have to be sorry about this, it's always good to have ideas and implementation to improve neru for everyone, and i appreciate it a lot.

PR #1093, opened 2026-07-25
chore(readme): update instructions to open the app after installed

keesiemeijer:
## Description
I've just started using this app, and it wasn't entirely clear that after the Homebrew install:

- The app was placed inside the Applications folder
- You needed to open the App before using it

Maybe this was explained in the terminal output,  but it can't hurt to put it in the readme as well.

This documentation change is Mac specific and is just a suggestion. I know too little about the app, and have no experience with Linux or Windows to also suggest changes for them.

I also had to do `neru config init` because the `config.toml` file wasn't created. I don't know if it should be created by default or if this is an extra step you have to do. This is maybe also something that could be added to the basic install instructions.

Thanks for the great app

## Target Platform

- [x] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix
- [x] `docs` — Documentation only

## Cross-Platform Checklist

## General Checklist

- [ ] Code formatted (`just fmt`)
- [ ] Linters pass (`just lint`)
- [ ] Tests pass (`just test`)
- [ ] Build succeeds (`just build`)
- [ ] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ brew tap y3owk1n/tap
 brew install --cask y3owk1n/tap/neru
 ```
 
-Open the Neru app in the Application folder and grant **Accessibility** in **System Settings → Privacy & Security → Accessibility**, then:
+Open the Neru app in the Applications folder and grant **Accessibility** in **System Settings → Privacy & Security → Accessibility**, then:
 
 Press `Cmd+Shift+Space`. Labels appear. You're already using it.
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ brew tap y3owk1n/tap
 brew install --cask y3owk1n/tap/neru
 ```
 
-Grant **Accessibility** in **System Settings → Privacy & Security → Accessibility**, then:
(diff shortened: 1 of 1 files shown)


y3owk1n:
Thanks for the PR, make sense. Haha. The app definitely have to be ran to start after installing. 

`config init` is a terminal command for terminal users. Just opening the app it self will have an onboarding process to do the config initialisation or just test it out without creating a config. 

This includes accessibility permission prompts too, we have a series of alert UI to guide you through the process. 

Anyways looks good to me, thanks!

keesiemeijer:
That makes sense. Maybe I couldn't find the file because I didn't open the app yet!

y3owk1n:
> That makes sense. Maybe I couldn't find the file because I didn't open the app yet! 

No worries, you can try to remove the created config from init command, and feel the process if you're interested with it. 😉

Will merge when everything is green. It's a good addition to the readme.

PR #1120, opened 2026-07-30
feat(held_repeat): accelerate step distance while a key is held

### LINKED ISSUE #1104: Time-based acceleration for held_key_repeat (esp. move_mouse_relative)
`held_key_repeat` fires the bound action at a constant `interval` while a key is held (e.g. `move_mouse_relative`). That's fine for scroll/page actions, but for mouse movement a single fixed step forces a tradeoff: small enough to place the cursor precisely, or large enough to cross the screen in reasonable time — never both.

A ramp would fix this: start slow, speed up the longer the key stays held, same way OS-level key repeat or game analog-stick acceleration works.

Proposed config, scoped to `held_key_repeat` (or a new `[mouse_acceleration]` block if that's cleaner):

```toml
[held_key_repeat]
enabled = true
initial_delay = 200
interval = 20
accel_enabled = true
accel_ramp_ms = 500     # time held to reach max multiplier
accel_max_multiplier = 4  # step size scales up to 4x by ramp_ms
```

Each repeat tick would scale the action's own `--dx`/`--dy` (or step-based actions like scroll) by the current multiplier, computed from how long the key's been held rather than a fixed value.

Use case: keyboard-only mouse movement via `move_mouse_relative` bound to home-row keys (no trackpad/mouse). Currently the only way to move the cursor any real distance is to either set a large per-tap step (imprecise for small adjustments) or hold the key at a mediocre fixed repeat speed. Acceleration would let one binding cover both precise nudges and long traversals.

Happy to open a PR if the shape above looks reasonable — wanted to check design direction first.

veyxov:
Adds optional acceleration to `[held_repeat]`: a held key's `--dx`/`--dy` ramp linearly from 1x to `accel_max_multiplier` over `accel_ramp_ms`. `interval_ms` stays fixed, so only the per-tick distance grows — no timer rescheduling. `accel_targets` is an allowlist; `move_mouse_relative` is the only valid entry today, since scaling acts on `--dx`/`--dy` and no other repeating action accepts them.

```toml
[held_repeat]
accel_enabled = true
accel_ramp_ms = 500
accel_max_multiplier = 4.0
accel_targets = ["move_mouse_relative"]
```

Wired into both repeat paths — global hotkeys (`hotkeys.go`) and per-mode hotkeys (`modes/key_dispatch.go`) — since `move_mouse_relative` is commonly bound under `[scroll.hotkeys]`.

The ramp is timed from the first repeat rather than the key press: nothing moves during `initial_delay_ms`, so an earlier start would land tick one mid-ramp and lurch.

Also collapses `isHeldRepeatAction` and `hotkeyActionsRepeatWhileHeld` onto one shared parser instead of adding a third copy.

## Testing

`go test ./...`, `golangci-lint run` clean. Unit tests cover flag scaling (both `--dx=N` and `--dx N` forms), the multiplier curve, allowlist matching, and validation rejections. Manually tested on macOS with home-row `move_mouse_relative` bindings.

Closes #1104

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -669,9 +669,13 @@ type SmoothScrollConfig struct {
 
 // HeldRepeatConfig defines held-key repeat settings for scroll, page, and mouse-move actions.
 type HeldRepeatConfig struct {
-	Enabled      bool `json:"enabled"      toml:"enabled"`          // Master toggle for held-key repeat
-	InitialDelay int  `json:"initialDelay" toml:"initial_delay_ms"` // Delay before first repeat fires (ms)
-	Interval     int  `json:"interval"     toml:"interval_ms"`      // Interval between subsequent repeats (ms)
+	Enabled            bool     `json:"enabled"            toml:"enabled"`              // Master toggle for held-key repeat
+	InitialDelay       int      `json:"initialDelay"       toml:"initial_delay_ms"`     // Delay before first repeat fires (ms)
+	Interval           int      `json:"interval"           toml:"interval_ms"`          // Interval between subsequent repeats (ms)
+	AccelEnabled       bool     `json:"accelEnabled"       toml:"accel_enabled"`        // Ramp step distance up while the key stays held
+	AccelRampMs        int      `json:"accelRampMs"        toml:"accel_ramp_ms"`        // Hold time to reach accel_max_multiplier (ms)
+	AccelMaxMultiplier float64  `json:"accelMaxMultiplier" toml:"accel_max_multiplier"` // Step distance multiplier at full ramp
+	AccelTargets       []string `json:"accelTargets"       toml:"accel_targets"`        // Action names eligible for acceleration
 }
 
 // SystrayConfig defines system tray settings.
--- a/internal/domain/action/accel.go
+++ b/internal/domain/action/accel.go
@@ -1,0 +1,79 @@
+package action
+
+import (
+	"math"
+	"strconv"
+	"strings"
+)
+
+// ScaleDeltaFlags multiplies the --dx/--dy values in actionStr.
+//
+// Tokenizing with Fields rather than SplitArgs because this rejoins: SplitArgs
+// consumes the quotes it parses. Only move_mouse_relative is ever scaled and its
+// flags are integers, so there is nothing to quote.
+func ScaleDeltaFlags(actionStr string, multiplier float64) string {
+	if multiplier == 1 {
+		return actionStr
+	}
+
+	tokens := strings.Fields(actionStr)
+
+	for idx, token := range tokens {
+		flag, value, joined := strings.Cut(token, "=")
+		if flag != "--dx" && flag != "--dy" {
+			continue
+		}
+
+		target := idx
+
+		if !joined {
+			if idx+1 >= len(tokens) {
+				continue
+			}
+
+			target, value = idx+1, tokens[idx+1]
+		}
+
+		parsed, err := strconv.Atoi(value)
+		if err != nil {
+			continue
+		}
+
+		scaled := strconv.Itoa(clampToInt(math.Round(float64(parsed) * multiplier)))
+		if joined {
+			scaled = flag + "=" + scaled
+		}
+
+		tokens[target] = scaled
+	}
+
+	return strings.Join(tokens, " ")
+}
+
+// Out-of-range float to int is implementation-defined in Go, so saturate.
+func clampToInt(value float64) int {
+	if value >= math.MaxInt {
+		return math.MaxInt
+	}
+
+	if value <= math.MinInt {
+		return math.MinInt
+	}
+
+	return int(value)
+}
+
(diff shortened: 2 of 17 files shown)


internal/core/domain/action/accel.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+)
+
+// HeldRepeatActionName returns the name of a lone "action <name> …" binding, or "".
+func HeldRepeatActionName(actions []string) string {
+	if len(actions) != 1 {
+		return ""
+	}
+
+	parts := strings.Fields(strings.TrimSpace(actions[0]))
```
y3owk1n: in `hotkeys.go` we have a split args function that handle quotes, i think we should use that if possible so that we have a consistent tokenisation. If not, i think we might have some edge cases that one part is handled but another rejected or something.

internal/config/validators.go
```diff
@@ -877,6 +877,34 @@ func (c *Config) ValidateHeldRepeat() error {
...
+			fmt.Sprintf(
+				"held_repeat.accel_max_multiplier must be between 1 and %d",
+				MaxHeldRepeatAccelMultiplier,
+			),
+		)
+	}
+
+	// Acceleration scales --dx/--dy, which only move_mouse_relative accepts.
+	// Anything else would validate and then accelerate nothing.
+	for _, target := range c.HeldRepeat.AccelTarget {
```
y3owk1n: i think we should also reject if the target is `[]` and is enabled.

internal/domain/action/accel.go
```diff
@@ -0,0 +1,78 @@
...
+
+		tokens[target] = scaled
+	}
+
+	return strings.Join(tokens, " ")
+}
+
+// ScaleAllDeltaFlags scales each action into a new slice, so repeated ticks
+// never compound onto the original binding.
+func ScaleAllDeltaFlags(actions []string, multiplier float64) []string {
```
y3owk1n, line 68: i think this could have a test for it

internal/config/config.go
```diff
@@ -930,9 +932,42 @@ type SmoothScrollConfig struct {
...
-	Enabled      bool `json:"enabled"      toml:"enabled"`          // Master toggle for held-key repeat
-	InitialDelay int  `json:"initialDelay" toml:"initial_delay_ms"` // Delay before first repeat fires (ms)
-	Interval     int  `json:"interval"     toml:"interval_ms"`      // Interval between subsequent repeats (ms)
+	Enabled            bool     `json:"enabled"            toml:"enabled"`              // Master toggle for held-key repeat
+	InitialDelay       int      `json:"initialDelay"       toml:"initial_delay_ms"`     // Delay before first repeat fires (ms)
+	Interval           int      `json:"interval"           toml:"interval_ms"`          // Interval between subsequent repeats (ms)
+	AccelEnabled       bool     `json:"accelEnabled"       toml:"accel_enabled"`        // Ramp step distance up while the key stays held
+	AccelRampMs        int      `json:"accelRampMs"        toml:"accel_ramp_ms"`        // Hold time to reach accel_max_multiplier (ms)
+	AccelMaxMultiplier float64  `json:"accelMaxMultiplier" toml:"accel_max_multiplier"` // Step distance multiplier at full ramp
+	AccelTarget        []string `json:"accelTarget"        toml:"accel_target"`
```
y3owk1n: actually, i think this should not be singular, but `accel_targets`, since the shape is an array.

y3owk1n (changes requested):

y3owk1n:
Thanks for the PR! I've added some comments, once resolved, we will move on to ensure our CI passes, and run greptile to catch bug alongside it.

Also i think linear curve is fine too at this moment, we can keep it for now.

internal/config/validators.go
veyxov: Done — rejects `accel_targets = []` when `accel_enabled = true`, with a test case. Same reasoning as the target validation right below it: it would otherwise validate and then accelerate nothing.

internal/domain/action/accel.go
veyxov: Done — added `TestScaleAllDeltaFlags` covering multi-entry slices, entries without delta flags mixed in, the `multiplier == 1` identity path, and the empty slice. Kept the existing aliasing test.

internal/config/config.go
veyxov: Done — renamed across config, defaults... and the docs.

veyxov:
All four comments addressed.I dropped my copy of the tokenizer and moved `HeldRepeatActionName` into `internal/config/held_repeat.go` to use yours.

y3owk1n:
@veyxov i am really sorry, but i have landed a huge refactoring across the code base for improvement on handling cross platform and navigating around the codebase. Will you be able to rebase from main? Sorry!!!!

veyxov:
No worries, rebased.

y3owk1n:
Looks good! I'll run CI to ensure the code are expected.

y3owk1n:
All good, thanks for the PR!

y3owk1n (approved):

PR #1135, opened 2026-08-01
refactor: make the cross-platform contract explicit and enforced

y3owk1n:
## Description

This PR reworks how Neru expresses platform support, so that adding a backend is
a matter of filling in a known slot rather than copying whichever nearby file
looks closest. Platform capabilities were previously expressed three different
ways — none of them named — and the inconsistency had already produced
contracts that were declared but never used, and a capability list that had
drifted out of sync with itself.

Every OS capability now belongs to one of three tiers, documented in the
contributor guide: a port, in-package build-tagged dispatch, or an optional port
extension. Applying that model uniformly removed two contracts nothing
implemented, replaced a bypassed one with the code that actually ships, and
turned four package-level functions the app was calling across layers into
proper ports. Moving the overlay backends into the infrastructure layer exposed
that they depended on application packages for their render models; those
packages moved down as well, so nothing in the core now depends on the
application layer.

Behaviour is unchanged except for `neru doctor`, which reports four additional
capabilities and no longer omits three of them on Windows. A set of guardrail
tests now enforces the rules that this PR establishes, so they hold without
relying on review to catch drift.

## Related Issues

None.

## Target Platform

- [x] macOS
- [x] Linux
- [x] Windows

## Type of Change

## Command changes

`neru doctor` and the IPC status/health responses now report four capabilities
that previously existed in code but were never surfaced, and the Windows
client-side output no longer drops three it should always have shown.

| Capability key | Reported as | Notes |
| -------------- | ----------- | ----- |
| `text_input` | `supported` on macOS, `stub` elsewhere | native hint-search field |
| `vision` | `supported` on macOS, `stub` elsewhere | OCR element detection |
| `key_feed` | `supported` on macOS and Linux, `stub` on Windows | synthetic key injection |
| `systray` | `supported` on all three | tray icon and menu |

Previously missing from the Windows client-side `neru doctor` output only, and
now included: `notifications`, `app_watcher`, `dark_mode_detection`.

No configuration options, commands, subcommands, flags or environment variables
were added, renamed or removed. Existing config files keep working untouched.

## Potential breaking changes

**Anything building Neru outside `just` or the Nix package must update its
`-ldflags`.** Version stamping moved out of the CLI package so that the tray
menu no longer has to reach into the command layer for the version string. The
`just` recipes, `nix/package.nix` and `docs/DEVELOPMENT.md` are updated in this
PR; a bespoke build script is not.

Before:

```
-X github.com/y3owk1n/neru/internal/cli.Version=$VERSION
-X github.com/y3owk1n/neru/internal/cli.GitCommit=$COMMIT
-X github.com/y3owk1n/neru/internal/cli.BuildDate=$DATE
```

After:

```
-X github.com/y3owk1n/neru/internal/buildinfo.Version=$VERSION
-X github.com/y3owk1n/neru/internal/buildinfo.GitCommit=$COMMIT
-X github.com/y3owk1n/neru/internal/buildinfo.BuildDate=$DATE
```

A build that misses this still compiles and runs; it silently reports version
`dev` instead of failing, which is worth knowing when reviewing release
tooling.

Arguably breaking, arguably not, and worth a second opinion: anything parsing
`neru doctor` or the IPC status payload now sees additional capability keys.
Nothing was removed or renamed, so a consumer reading specific keys is
unaffected, but one asserting on the exact key set is not.

## Cross-Platform Checklist

## General Checklist

## Additional Context

**Guardrail tests.** `internal/architecture` now enforces the tier model,
dependency direction, file-slot naming, cgo suffix naming, package-comment
reachability and port/mock coverage. Each check was verified against a
deliberately planted violation rather than merely observed to pass. Two of the
checks keep their own exception lists honest by failing when an entry stops
being a real violation, so the lists can only shrink. The layering exception
list is currently empty.

**Bugs found while applying the model.** The `internal/core/infra/textinput`
package could not build a test binary on macOS at all, because it never linked
the Objective-C bridge it calls — which is why it had no tests. The
`platform/linux` package comment lived in a `linux && cgo` file, so `CGO_ENABLED=0`
builds had no package documentation; three overlay render packages had the same
problem per-OS. The no-cgo uinput stubs returned bare errors where the
documented contract calls for `CodeNotSupported`, defeating
`derrors.IsNotSupported` for any caller that checks it.

**Tests.** Every test that asserted nothing was either given real assertions or
removed. Five `*ImplementsPort` tests were deleted as theatre — each duplicated
a package-scope `var _ ports.X` assertion that already fails at compile time.
Seven tests took `_ *testing.T` and so were structurally incapable of failing.
The strengthened tests covering the largest behavioural surfaces were
mutation-tested: breaking the code under test was confirmed to fail them.

**Verification.** `just fmt && just lint && just test && just build` is green, as
is `just test-race`. `GOOS=linux CGO_ENABLED=0` and `GOOS=windows` cross-builds
succeed and the integration-tagged suite compiles. The cgo Linux paths could not
be exercised on the macOS host used here — this PR rewrites cgo include paths
across two package moves and splits a `linux`-tagged file five ways, so the
`ubuntu-latest` CI job is the meaningful check on those.

**Reviewing this.** The diff is large and mostly mechanical. The file splits were
verified symbol-by-symbol: every top-level declaration in each original file is
accounted for in the resulting files, with none lost. The file worth the closest
read is `internal/app/app_initialization_steps.go`, which was reconstructed by
hand partway through the work; it builds and passes, but it saw manual
reassembly the other files did not.


y3owk1n:
@greptile-apps

PR #1139, opened 2026-08-02
feat(action): add `run` for ordered action sequences

y3owk1n:
## Description

This PR adds a way to run several actions in order as one unit, from a single command. `neru run "action save_cursor_pos" "hints --action left_click"` executes the steps in the daemon, so a script or an external hotkey daemon no longer has to spawn one `neru` process per step and hope the pieces line up.

Underneath, the same thing was implemented three separate times — for global hotkeys, for per-mode hotkeys, and for comma-separated click chains — with rules that had already begun to diverge. They now share one implementation, so a sequence behaves identically whether it is written in a hotkey binding, passed to a mode's `--on-exit`, or run from a script. As a result `--on-exit` can now take several steps instead of one, which collapses the multi-binding workaround that recipes needed for "click, wait, then do something else".

Two smaller fixes came out of the same area. A command that took longer than the connection's 30-second budget had its reply refused before it was attempted, so the caller saw a timeout for work that had actually completed; the read and write sides are now budgeted separately. And a `sleep` step ignored daemon shutdown, holding a goroutine until its duration elapsed.

This PR adds no new expressiveness beyond ordering — no conditionals, no variables, no retries. Failure semantics stay as they are today: a bail stops the sequence, any other failure is reported and the following steps still run.

## Related Issues

None.

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [x] `feat` — New feature

**New command**

| Command | Description |
| --- | --- |
| `neru run <step> [step...]` | Runs the given steps in order. Each argument is one step in the hotkey action grammar: an action (`action left_click`), a mode (`hints --action left_click`), or a shell command (`exec open -a Safari`). |

```bash
neru run "action save_cursor_pos" "hints --action left_click" \
         "action wait_for_mode_exit --bail" "action restore_cursor_pos"
```

Exit status is `0` on success, `ERR_CHAIN_BAIL` when a step stopped the sequence, and `ERR_ACTION_FAILED` naming the first failing step otherwise.

**Changed flag**

| Flag | Before | Now |
| --- | --- | --- |
| `--on-exit` on `hints`, `grid`, `recursive_grid` | Single value | Repeatable; each occurrence appends one step, run in order |

```toml
[hotkeys]
"Primary+Shift+Space" = [
    "action save_cursor_pos",
    "hints --action left_click --on-exit 'action restore_cursor_pos'",
]
```

Existing configs keep working untouched: a single `--on-exit` behaves exactly as before, and it still runs only when the mode's action is fulfilled — never on escape or `neru idle`.

**New accepted value**

`run` is now valid as a hotkey binding action, alongside `hints`, `idle`, `exec` and the rest. It was previously rejected by config validation as an unknown command. Arrays remain the idiomatic spelling inside a config; this is for consistency, so a step that works in one place works everywhere.

## Potentially breaking

Two behaviour changes that a very unusual setup could notice. Neither has a migration; both are judgement calls worth a second opinion.

- **Repeated `--on-exit` in one mode invocation.** Previously the last occurrence won and the earlier ones were discarded; now all of them run, in order. Only affects a binding that already passes the flag more than once, which had no reason to exist since the extra values were silently dropped.
- **Held-key repeat and bail.** A held-repeat binding whose sequence bails now stops that tick's remaining steps instead of running them. Each tick still starts fresh, so repeating continues. This aligns held repeat with the other two dispatch paths, which already stopped on a bail — arguably a bug fix rather than a break, but it is a behaviour change in shipped code.

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no platform-specific code in this change

## General Checklist

## Additional Context

**Why one executor.** The three implementations disagreed already: held-key repeat ran every step of a binding even after one asked the chain to stop, while the other two paths stopped. Consolidating fixes that by construction and makes the queued follow-ups — named macros, an explicit `--bail-on-error` — one change each instead of three with three chances to drift.

**Recursion.** Because a step can itself be a `run`, a binding can refer back to itself. Nesting depth travels in the step context and is refused past five levels. This is reachable in practice through `neru run "run ..."` and through `--on-exit 'run ...'`, neither of which passes through config validation.

**Connection deadlines.** The 30-second deadline set at accept was being spent by handler execution, which is not I/O. The read side keeps that budget, which is what guards a client that connects and never sends a command; the write side gets its own five-second window starting after the handler returns. The alternative considered — classifying the resulting `deadline exceeded` as a benign "peer gone" error — was rejected because it would have hidden dropped replies rather than delivered them, and would have made a genuinely stuck write look benign forever.

**Verification beyond the test suite.** Driven against a running daemon started with an empty `[hotkeys]` table: a 3-step sequence, a bail stopping the remaining steps, a failing step not stopping them, `run` nested one and two levels deep, `--on-exit` not firing on a cancelled mode and firing in order on a fulfilled one, and a 9-step sequence combining mode activation, fed keys, a blocking wait and cursor save/restore. A 37-second sequence returned its result normally, where it previously failed with a client timeout. `SIGTERM` during an in-flight `action sleep 60s` exited the daemon immediately and released the step. Race detector clean across the touched packages.

**Measured overhead.** Five steps as five separate invocations took 30–40 ms; the same five as one `neru run` took under 10 ms — roughly 6–7 ms of process spawn and connection setup per step, removed for every step after the first.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -21,6 +21,7 @@ The same content is available as manpages (`man neru`) after installation.
 - [Daemon lifecycle](#daemon-lifecycle) — `launch` · `start` · `stop` · `idle` · `status` · `doctor`
 - [Navigation modes](#navigation-modes) — `hints` · `grid` · `recursive_grid` · `scroll` · `monitor_select`
 - [Actions](#actions) — `action` and its subcommands
+- [Sequences](#sequences) — `run`
 - [Configuration commands](#configuration-commands) — `config`
 - [Runtime toggles](#runtime-toggles)
 - [Utilities](#utilities) — `roles` · `services` · `docs`
@@ -88,6 +89,7 @@ Accepted by every command.
 | [`scroll`](#neru-scroll)                                     | Vim-style scrolling             | Yes | All |
 | [`monitor_select`](#neru-monitor_select)                     | Jump the cursor to a display    | Yes | macOS · Linux |
 | [`action`](#actions)                                         | One-shot mouse/scroll/key input | Yes | All ² |
+| [`run`](#neru-run)                                           | Run several actions in order    | Yes | All |
 | [`config`](#configuration-commands)                          | Inspect and change config       | Mixed | All |
 | [`toggle-scroll-invert`](#neru-toggle-scroll-invert)         | Invert scroll direction         | Yes | All |
 | [`toggle-cursor-follow-selection`](#neru-toggle-cursor-follow-selection) | Toggle cursor follow | Yes | All |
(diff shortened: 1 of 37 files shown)


y3owk1n:
Fixed. `flattenRunSteps` now recurses to `maxSequenceDepth` instead of expanding a single level, so a mode named below the first `run` is visible to both the mode-switch modifier suppression and the disabled-mode check.

The comment justifying one level was also wrong — it claimed the callers only need to know whether a mode is named *anywhere* in the binding, which one-level expansion does not deliver. Expansion now stops exactly where the executor does: past the depth limit a sequence is refused, so its steps never run and the unexpanded `run ...` correctly names no mode.

Covered by two new cases in `TestBindingInspectionSeesModesInsideRun` (mode inside a nested run, nested run without a mode) and `TestFlattenRunSteps_StopsAtTheExecutorsNestingLimit` for the boundary. Verified the nested case fails against the previous one-level implementation. Amended into the existing commit; gate is green.

PR #1141, opened 2026-08-02
feat(config): add `[macros]` for named, reusable action sequences

y3owk1n:
## Description

This PR adds named action sequences, so a sequence used by several keys is
written once instead of copied across them.

Previously the only way to share steps between bindings was to repeat them. The
per-app recipe in the tips guide for switching between views in one window
repeats the same five steps three times over, with only two numbers differing —
that recipe is now three one-line bindings and one definition.

A macro is invoked with `macro <name> [args...]` from anywhere a sequence step
is accepted: a hotkey binding, a mode's `--on-exit`, `neru run`, or another
macro. Arguments are positional. Calls are checked when the config loads rather
than when the key is pressed, because a binding runs in the background and
discards its result — a mistyped name would otherwise make the key appear to do
nothing at all.

This adds naming and argument substitution, and nothing else: no conditionals,
no state, no recording. A macro body is the same list of steps you would have
inlined into the binding.

Making the load-time check trustworthy meant giving validation the same reach
execution already has. That closed three surfaces which previously accepted any
action string at all — `monitor_select` bindings, both Mission Control hooks,
and every step nested inside a `run` or an `--on-exit`. Configurations that were
silently carrying a broken step there will now be reported; see
[Potentially breaking](#potentially-breaking).

## Related Issues

None.

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [x] `feat` — New feature

**New config table**

| Table | Description |
| --- | --- |
| `[macros]` | Named action sequences. Each key is a macro name, each value a step or list of steps. Optional; omitting it changes nothing. |

```toml
[macros]
click_and_exit = ["action left_click --bail-on-error", "idle"]
window_click = [
    "action move_mouse --window --x -1000 --y -1000",
    "action sleep 0.1",
    "action move_mouse_relative --dx $1 --dy $2",
    "action left_click",
]

[hints.hotkeys]
"Enter" = "macro click_and_exit"

[[app_configs]]
bundle_id = "com.example.someapp"
hotkeys = { "Cmd+1" = "macro window_click 100 70", "Cmd+2" = "macro window_click 250 70" }
```

**New step keyword**

| Step | Where | Description |
| --- | --- | --- |
| `macro <name> [args...]` | Any sequence: `[hotkeys]`, a per-mode table, a per-app override, `--on-exit`, `neru run`, another macro body | Runs the named sequence. |

**Rules a user needs to know**

- **Names** use letters, digits, `_` and `-`, and start with a letter.
- **Arguments** are positional: `$1`, `$2`, …, with `$$` for a literal dollar.
  Substitution is textual and happens before the step is split into arguments,
  so quote a placeholder that may contain spaces — `exec say "$1"` — as in a
  shell.
- **Arity is exact**, and checked at load: a call must pass as many arguments as
  the body's highest placeholder. An unknown name or the wrong count fails
  `neru config validate`.
- **A placeholder cannot be the command word.** `"$1 --action left_click"` as a
  body step is rejected at load, since a step whose command is known only at
  call time could not be validated at all.
- **A macro is not accepted as a mode's `--action`**, which takes a mouse-button
  name; use `--on-exit` for a sequence that follows the action.
- **A macro runs as a nested sequence**, so a failure inside it reaches the
  caller as that one step failing, and a `--bail-on-error` or `--stop-on-error`
  on the caller acts on it. The caller's policy does not apply inside the body.
- **A macro may call another macro**, and one that reaches itself is stopped by
  the same five-level nesting limit that bounds `run`.
- **Calls are checked wherever they are written**: `[hotkeys]`, every
  `[<mode>.hotkeys]` table, the per-app overrides of both,
  `hints.on_mission_control_activated` / `_deactivated`, a macro body, and the
  steps carried inside a `run` or an `--on-exit`.

Existing configs are unaffected: without a `[macros]` table nothing changes, and
a config written before this still loads and validates unchanged.

## Potentially breaking

Validation now reaches surfaces it did not before, so a configuration that
loads today can be reported as invalid after this change. **An invalid
configuration stops the daemon from starting**, so this is worth a deliberate
call rather than a footnote.

Three surfaces previously accepted any action string, checking nothing:

| Surface | Before | Now |
| --- | --- | --- |
| `[monitor_select.hotkeys]` | any string accepted | checked like every other mode table |
| `hints.on_mission_control_activated` / `_deactivated` | macro calls unchecked | checked like every other binding |
| Steps inside a `run` or an `--on-exit` | any string accepted | checked like a step written directly |

Who this affects: someone whose config contains a step that never worked — a
typo such as `run 'action lefft_click'`, or a `[monitor_select.hotkeys]` entry
naming an action that does not exist. That binding is already dead at runtime;
the change is that Neru now says so at load instead of failing silently when
the key is pressed.

What they do: run `neru config validate`, which names the offending field and
step, and fix or remove it. Every config shipped in `configs/`, including the
maintainer's own, validates unchanged.

The alternative — warning for one release and failing afterwards — would avoid
the upgrade cliff at the cost of two classes of validation error to explain.
I went with failing, because `[hints.hotkeys]` has always failed on a bogus
action and the surfaces that did not were the accident. Happy to switch it to a
warning if you would rather stage it.

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no platform-specific code in this change

## General Checklist

- [x] Tests pass (`just test`) — unit suite and the integration tests for the packages this touches; see Additional Context for the one suite that cannot run locally right now

## Additional Context

**On the name.** "Macro" carries a record-and-replay connotation from text
editors, which this is not; the docs say so where a reader will hit it. It was
kept over "workflow" — already used throughout the docs in its ordinary prose
sense, and suggestive of stages and state this does not have — and over
"sequence", which is the codebase's own word for the generic concept. The
parameterised-substitution sense of "macro" describes the implementation
exactly, and it is what comparable tools in this space call the same thing.

**Why nested rather than inlined.** Splicing a body into its caller would have
hidden the nesting from the depth guard, so a self-referential macro would
recurse freely, and would have reported failures against an expanded step index
the author never wrote. Running it as a nested sequence gets both for free.

**Why the tokenizer moved.** Validation has to read a call exactly as the daemon
dispatches one, or the two disagree about where an argument begins — the same
divergence the shared executor removed earlier. There is now one definition,
in the package that owns the config format, and the app delegates to it.

**Binding inspection.** The checks that decide modifier suppression and
disabled-mode skipping have to see a mode named inside a macro, or a binding
written as a macro would silently skip them. They now walk into bodies with
early return instead of expanding into a slice, and the walk is bounded: macros
that fan out into each other multiply at every level, and this runs on the
key-press path. Past the budget the answer is "no mode named here", the same
conclusion already drawn for anything that cannot be expanded.

**Known and deliberate.** A macro's steps can fan out at execution time — depth
is bounded, width is not — in the same way a very long hotkey array can; the
config does what it says. A macro that reaches itself reports one `macro "x":`
wrapper per level before the real cause, which is verbose but shows the chain.

**On the review that produced the validation work.** An automated review
flagged macro calls escaping validation in "Mission Control hooks". I checked,
found nothing by that name, and said so — wrongly: the grep I based that on was
truncated, and `hints.on_mission_control_activated` / `_deactivated` are real
action sequences dispatched through the same executor. The second flag was
correct. Rather than patch just that hook, the walk was extended to every
surface and every nesting level, which is what turned up `monitor_select` and
nested steps as well.

**Verification beyond the test suite.** Driven against a running daemon:
arguments substituted, a nested macro, an unknown name, the wrong argument
count, a self-invoking macro stopped by the depth guard, a failing body step
tolerated by default, the same step made fatal with `--bail-on-error` on the
call, `--stop-on-error` at the call site, and `--bail-on-error` correctly not
counted as one of the macro's arguments. A macro added after startup is picked
up by `neru config reload`, and `[macros]` survives a `neru config set` rewrite
intact. Race detector clean; no panics.

Each newly covered surface was probed before and after: a bogus action in
`[monitor_select.hotkeys]`, in either Mission Control hook, inside a `run`, and
inside an `--on-exit` all loaded clean before and are reported now, while the
valid form of each still passes. Two tests pin this — one that every surface
rejects a bad step, and one that every surface still accepts a good one, so the
coverage cannot pass by rejecting everything.

**One suite could not run locally.** `TestHintServiceIntegration` performs an
unbounded real accessibility scan of the focused window and hangs while the
screen is locked, which is where this machine was. It is unrelated to this
change — reproduced at `02c01dae` in a clean worktree — but it means the full
`just test` was not observed green end to end here; the unit suite and the
integration tests for `config`, `app` and `ipc` were.

### DIFF
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -45,6 +45,17 @@ text = "#E8EEFF"
 "Primary+Shift+C" = "recursive_grid"
 "Primary+Shift+S" = "scroll"
 
+# Macros — named action sequences, reused from any binding.
+# Written by hand, not recorded. $1, $2 ... are the call's arguments.
+# See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#macros
+#
+# [macros]
+# click_and_exit = ["action left_click --bail-on-error", "idle"]
+# click_at = ["action move_mouse --x $1 --y $2", "action left_click"]
+#
+# [hints.hotkeys]
+# "Enter" = "macro click_and_exit"
+
 # Hint mode
 # See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#hints
 [hints]
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -876,8 +876,9 @@ neru run <step> [step...]
 
 Each argument is one step, written exactly as it would be written in a hotkey
 binding: an action (`action left_click`), a mode (`hints --action left_click`),
-or a shell command (`exec open -a Safari`). The daemon executes the steps in
-order.
+a shell command (`exec open -a Safari`), or a named sequence from the
+[`[macros]`](CONFIGURATION.md#macros) table (`macro window_click 100 70`). The
+daemon executes the steps in order.
 
 This is the same executor that runs a multi-action hotkey binding, so a
 sequence behaves identically whether it is written in `[hotkeys]`, passed to
@@ -929,6 +930,11 @@ neru run --stop-on-error "action left_click" "action restore_cursor_pos"
 It works in any sequence — a hotkey binding, a mode's `--on-exit`, or `neru run`.
 Text that merely looks like the directive is left alone, so
 `exec sh -c "echo --bail-on-error"` still passes it through to the shell.
+
+The policy applies to the steps of one sequence. A step that runs a nested
+sequence — another `run`, or a [`macro`](CONFIGURATION.md#macros) — keeps its
+own policy inside; its overall failure is then reported to the caller as that
+one step failing, which an outer `--stop-on-error` or `--bail-on-error` acts on.
 
 A stopped sequence reports which step ended it and that the later steps did not
 run; a tolerated failure says the opposite, so the two are distinguishable by a
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -30,6 +30,7 @@ keeps its default. "The daemon" below means the process started by
 
 | Section                                       | Controls                                     |
 | --------------------------------------------- | -------------------------------------------- |
+| [`[macros]`](#macros)                         | Named action sequences reused across bindings |
 | [`[general]`](#general)                       | Global behaviour, passthrough, `exec` shell   |
 | [`[theme]`](#theme)                           | Base palette all components derive from       |
 | [`[hints]`](#hints)                           | Hints mode and element discovery              |
@@ -564,6 +565,70 @@ An array like the ones above is an *action sequence*. The same sequence, with th
 ---
 
 # Sections
+
+## [macros]
+
+Named action sequences. A macro is written once and invoked from any binding
+with `macro <name> [args...]`, which keeps a sequence used by several keys in
+one place instead of copied across them.
+
+These are written, not recorded: there is nothing to capture and replay, unlike
+the macros of a text editor. A macro is a named list of the same steps you would
+otherwise inline into the binding, with positional arguments.
+
+```toml
+[macros]
+# No arguments.
+click_and_exit = ["action left_click --bail-on-error", "idle"]
+
+# $1 and $2 are the first and second argument of the call.
+window_click = [
+    "action move_mouse --window --x -1000 --y -1000",
+    "action sleep 0.1",
+    "action move_mouse_relative --dx $1 --dy $2",
+    "action left_click",
+]
+
+[hints.hotkeys]
+"Enter" = "macro click_and_exit"
+
+[[app_configs]]
+bundle_id = "com.anthropic.claudefordesktop"
+hotkeys = { "Cmd+1" = "macro window_click 100 70" }
+```
+
+**Names** use letters, digits, `_` and `-`, and start with a letter.
+
+**Arguments** are positional: `$1`, `$2`, and so on, with `$$` for a literal
+dollar sign. Substitution is textual and happens before the step is split into
+arguments, so quote a placeholder that may contain spaces — `exec say "$1"` —
+exactly as you would in a shell.
+
+**Arity is checked when the config loads.** A call must pass exactly as many
+arguments as the body uses; an unknown name or the wrong count fails
+`neru config validate` rather than doing nothing when the key is pressed.
+
+The check reaches everywhere an action can be written: `[hotkeys]`, every
+`[<mode>.hotkeys]` table, the per-app overrides of both, the
+[Mission Control hooks](#hints), a macro body, and the steps carried inside a
+`run` or an `--on-exit`. So does the ordinary check that a step names a real
+command — a step is validated at whatever depth it appears.
+
+**A placeholder belongs in an argument, not in the command word.** A body step
+like `"$1 --action left_click"` is rejected at load, because a step whose
+command is only known at call time could not be validated at all.
+
+**A macro body is a sequence** like any other, so it can use
+[`--bail-on-error`](CLI.md#failure-policy), and it can call another macro. A
+macro that reaches itself is stopped by the same nesting limit that bounds
+`run` — five levels — rather than recursing.
+
+**A macro runs as a nested sequence**, so a failure inside it is reported to the
+caller as that one step failing. Mark the call with `--bail-on-error` when the
+rest of the calling sequence should not continue past it.
+
+Macros are not accepted as a mode's `--action`, which takes a mouse button
+name; use `--on-exit` for a sequence that follows the action.
 
 ## [general]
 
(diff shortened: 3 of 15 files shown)


y3owk1n:
**Issue 1 — valid, fixed.** A macro invoked from inside a `run` step, or from a mode's `--on-exit`, was never reached by the validation walk, so it loaded clean and failed only when the binding ran. That defeats the reason the check exists at load in the first place, since a hotkey dispatches in the background and discards its result.

Validation now descends into the steps a step carries — `run` arguments and `--on-exit` values, in both the separated and `=` spellings — bounded by the same nesting limit the runtime uses. Verified against a real config before and after:

```
run 'macro typo'                                  before: valid   after: no macro named "typo"
run 'run "macro typo"'                            before: valid   after: no macro named "typo"
hints --action left_click --on-exit 'macro typo'  before: valid   after: no macro named "typo"
run 'macro real'                                  before: valid   after: valid
```

Five cases added to `TestValidateMacros` (nested one and two levels, `--on-exit`, wrong arity nested, and a valid nested call so the fix cannot just reject everything). I also made `--on-exit` a single shared constant so the config validator and the IPC parser cannot drift on its spelling.

The "Mission Control action" part I could not substantiate: the only occurrences of Mission Control in the tree are in the darwin event tap, and there is no config surface by that name. Every table that does hold action strings is walked — global bindings, the five per-mode tables, the root and four per-mode `app_configs`, and the macro bodies. `monitor_select` has no `app_configs` to walk.

**Issue 2 — not a defect, declined, but now documented in the code.** Reading the macro table per call is deliberate and matches how every other step reads configuration: `action scroll_down` takes the step size current when it fires, not one pinned at the start of the sequence. Pinning would trade this for the opposite problem — a sequence that waits minutes for a mode to exit would then run definitions the user had already edited.

On the specific risks named:

- *Mixing old and new bodies.* One body is read once and expanded once, so a single invocation is always internally consistent. Across two invocations you can get old-then-new, exactly as two steps reading any other config value can. There is no torn read: reload swaps the whole config pointer and the map is never mutated in place.
- *Failing after partial execution.* If a reload deletes a macro mid-sequence, the later call fails with `no macro named "x"`. That is the correct outcome — the alternative is silently running a definition the configuration no longer contains.

The window is also a reload landing inside a sequence's execution. Rather than change behaviour, I added a comment at the lookup site stating the read-at-dispatch policy and why, so the next reader does not have to re-derive it.

Amended into the existing commit; gate is green.

y3owk1n:
**You were right and I was wrong — twice, and I should own the first one properly.**

`hints.on_mission_control_activated` and `hints.on_mission_control_deactivated` are real action sequences, dispatched through the same executor as any binding. When I dismissed the first report I based it on a grep whose output was truncated, and I stated the conclusion — "there is no config surface by that name" — with more confidence than the evidence supported. The second flag was correct, and the check that settled it was an exhaustive enumeration of every string-list field in the config struct rather than a keyword search.

Reproduced before the fix:

```
on_mission_control_activated = ["macro typo"]   ->  Configuration is valid
```

**Fixed, and widened past what was reported.** Rather than add the two hooks and stop, I asked what else execution reaches that validation does not. Three surfaces accepted any action string at all:

| Surface | Before | Now |
| --- | --- | --- |
| `hints.on_mission_control_activated` / `_deactivated` | macro calls unchecked | checked |
| `[monitor_select.hotkeys]` | anything accepted — the table was simply missing from the mode list, while its four siblings were there | checked |
| Steps inside a `run` or an `--on-exit` | anything accepted | checked at any depth |

All six cases now report precisely, and the valid form of each still passes:

```
MC hook: unknown macro          -> no macro named "typo"
MC hook: bogus action           -> unknown action subcommand: bogus_thing
MC hook: valid macro            -> Configuration is valid
monitor_select: bogus action    -> unknown action subcommand: bogus_thing
run nested: bogus action        -> unknown action subcommand: bogus_thing
--on-exit: bogus action         -> unknown action subcommand: bogus_thing
```

Two tests pin it: one that every surface rejects a bad step, and one that every surface still accepts a good one, so the coverage cannot pass by rejecting everything. Every config in `configs/`, including the maintainer's own, validates unchanged.

This widening can report a configuration that loads today, and an invalid config stops the daemon from starting, so the PR body now carries a **Potentially breaking** section describing exactly who is affected and what they do about it.

**Issue 2 (nested macros re-reading the config) — still declined, unchanged from my previous reply.** Reading the table per call is how every step in the daemon reads configuration; a body is read once and expanded once, so a single invocation is internally consistent, and there is no torn read because reload swaps the whole config pointer and never mutates the map. Pinning would trade this for a sequence that waits minutes running definitions the user had already edited. The policy is now stated in a comment at the lookup site so it does not have to be re-derived.

PR #1142, opened 2026-08-02
fix(config): report an invalid reload instead of blocking on the alert

y3owk1n:
## Description

This PR fixes `neru config reload` reporting a timeout when the configuration is
invalid.

Reloading a config that fails validation showed a native alert from inside the
reload. On macOS that alert is modal, so the reload sat behind it until the
dialog was dismissed — and the CLI gave up first:

```
$ neru config reload
[IPC_FAILED] failed to send command: [TIMEOUT] receive timeout: neru may be unresponsive
```

The daemon was neither unresponsive nor wrong: it had already diagnosed the
problem precisely and kept the previous configuration. It just could not say so
until someone clicked a button. The alert now goes up alongside the reply
instead of in front of it, so the reload reports the validation error straight
away. The alert is still shown, because the systray's Reload item has no other
way to report a failure, and at most one alert is outstanding at a time.

This PR also fixes two things in the test suite that share a cause with the
reload — a local development loop reporting something other than what is
actually wrong. The darwin service integration tests get a watchdog, and the
example-config test stops treating every file in `configs/` as one the project
ships.

## Related Issues

None.

## Target Platform

- [x] macOS
- [ ] Linux
- [ ] Windows

The reload fix is platform-agnostic; the symptom is macOS-specific because that
is where the alert is modal. The test change is in a darwin-tagged file.

## Type of Change

- [x] `fix` — Bug fix

None. No option, command, flag, or default changes. `neru config reload` on an
invalid config now exits with the validation error it always had rather than a
timeout:

```
$ neru config reload
[IPC_FAILED] configuration validate configuration failed: … hints.hotkeys.x: unknown action subcommand: bogus_thing
```

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — the touched test file keeps its `integration && darwin` tags
- [x] N/A — This PR does not touch platform-specific code — the reload change is shared code; only a darwin-tagged test file is otherwise involved

## General Checklist

- [x] Tests pass (`just test`) — the full suite, integration included
- [x] Documentation updated (if applicable) — N/A, no user-facing interface changed

## Additional Context

**Why the alert stays.** Two things trigger a reload: the IPC path
(`neru config reload`, and `config set` / `config reset`, which reload to apply)
and the systray's Reload menu item. The IPC caller gets the error in its reply,
but the systray click has no reply to carry it, so removing the alert would make
that path fail silently. Detaching it from the reply keeps both honest. It runs
under a context derived with `context.WithoutCancel`, so completing the reply
cannot tear down a dialog the user is still reading.

**Why this surfaced now.** The previous change widened validation to cover
`monitor_select` bindings, both Mission Control hooks, and steps nested inside a
`run` or an `--on-exit`. That is a good change, but it means more reloads now
fail validation, and every one of them went through the blocking path. Traffic
through a door that sticks.

**The test change.** The darwin service integration tests drive live
accessibility APIs against whatever window is focused. When the screen is locked
the focused bundle is `com.apple.loginwindow` and that call does not return, so
`just test` stops there indefinitely with no output explaining why — I lost two
runs to it, and had to bisect against a clean worktree to establish it was
pre-existing rather than something I had introduced.

A context deadline cannot bound this, which is worth stating plainly because it
was my first attempt: several adapters take a `context.Context` and discard it
outright (`Scroll(_ context.Context, …)`), and those that do consult it check
only before starting work, so nothing interrupts a call already blocked inside
the platform API. Passing a deadline would have advertised a bound that does not
exist. A watchdog runs instead, one per subtest: if a subtest has not finished
within 30 seconds the run ends with a message naming the likely cause, and the
accompanying stack shows which call is stuck.

Per subtest rather than per test function, because a function holds several and
a shared budget would eventually blame a locked screen for what is only a slow
machine. Demonstrated with a 900ms budget and four subtests sleeping 400ms
each: shared, that panics with a false diagnosis; per subtest, it passes.
Forcing the budget to 1ns still ends the run with the diagnosis and a stack, so
a genuine hang is caught either way.

**On the accumulating alerts.** Showing the dialog inline had one property
worth keeping: it serialized failed reloads, since a second could not start
until the first was dismissed. Moving it off the reply's path removed that, so a
run of failures would have stacked dialogs and parked a goroutine behind each.
A single outstanding alert is now enforced; once dismissed, the next failure
alerts again rather than being suppressed for good.

**The example-config test.** It globbed `configs/*.toml`, but that directory is
also a working area: a config kept there for local testing was validated as
though the project published it, and failed the suite over a file that is not a
project artifact. The shipped examples are now listed explicitly, so the test
asserts what it means to assert — the configs users copy are valid. The tradeoff
is that adding a new example means adding it to that list, which is the
deliberate step that puts it under test.

**Verification.** A blocking alert provider stands in for the modal dialog in a
new test: the reload must return its error while the alert is still up, and the
alert must still be shown. Confirmed it fails against the previous inline
version — it hangs until the 5-second guard — and passes now. A second test
drives five failing reloads against an undismissed alert and asserts only one
was shown, then that dismissing it lets the next failure alert again; it fails
against the unguarded version. Race detector clean. The full `just test`,
integration suite included, is green with the screen unlocked.

### DIFF
--- a/internal/app/services/services_integration_darwin_test.go
+++ b/internal/app/services/services_integration_darwin_test.go
@@ -4,8 +4,10 @@ package services_test
 
 import (
 	"context"
+	"fmt"
 	"image"
 	"testing"
+	"time"
 
 	"go.uber.org/zap"
 
@@ -21,6 +23,49 @@ import (
 	"github.com/y3owk1n/neru/internal/core/infra/platform"
 	"github.com/y3owk1n/neru/internal/core/ports"
 )
+
+// realAdapterTimeout bounds how long these tests wait on the real system.
+//
+// They drive live accessibility APIs against whatever window happens to be
+// focused, and that call can take arbitrarily long — with the screen locked,
+// where the focused "app" is the login window, it does not return at all.
+const realAdapterTimeout = 30 * time.Second
+
+// startAdapterWatchdog ends the test run with a diagnosis if a real system
+// call never comes back. Each subtest starts its own: a test function holds
+// several, and one budget shared across them would eventually blame a locked
+// screen for what is only a slow machine. Setup outside a subtest is left to
+// the go test -timeout backstop.
+//
+// A context deadline cannot do this job: several adapters take a context and
+// discard it, and the ones that honor it check it only before starting work,
+// so nothing interrupts a call already blocked inside the platform API.
+// Panicking is what turns an indefinite hang into something a reader can act
+// on — the message names the likely cause, and the stack dump that comes with
+// it names the call that is stuck.
+func startAdapterWatchdog(t *testing.T) {
+	t.Helper()
+
+	finished := make(chan struct{})
+	t.Cleanup(func() { close(finished) })
+
+	name := t.Name()
+
+	go func() {
+		select {
+		case <-finished:
+		case <-time.After(realAdapterTimeout):
+			panic(fmt.Sprintf(
+				"%s: a real system adapter did not answer within %s. "+
+					"This usually means the screen is locked, or the focused app is not "+
+					"responding to accessibility queries, rather than a fault in the code "+
+					"under test. The stack below shows which call is blocked.",
+				name,
+				realAdapterTimeout,
+			))
+		}
+	}()
+}
 
 // testThemeProvider is a simple ThemeProvider mock for integration tests.
 type testThemeProvider struct {
@@ -69,6 +114,8 @@ func TestHintServiceIntegration(t *testing.T) {
 	ctx := context.Background()
 
 	t.Run("ShowHints integration", func(t *testing.T) {
+		startAdapterWatchdog(t)
+
 		// This tests the full pipeline: accessibility -> hint generation -> overlay.
 		// Zero hints is a legitimate outcome (the machine may have no clickable
 		// elements on screen), but an error is not.
@@ -103,6 +150,8 @@ func TestHintServiceIntegration(t *testing.T) {
 	})
 
 	t.Run("ShowHints leaves NoOpManager-backed overlay idle", func(t *testing.T) {
+		startAdapterWatchdog(t)
+
 		// initializeRealAdapters wires a NoOpManager whose Mode() is pinned to
 		// ModeIdle, and Adapter.IsVisible is defined as Mode() != idle. So the
 		// adapter must report not-visible no matter what was shown. If this
@@ -118,6 +167,8 @@ func TestHintServiceIntegration(t *testing.T) {
 	})
 
 	t.Run("HideHints integration", func(t *testing.T) {
+		startAdapterWatchdog(t)
+
 		err := hintService.HideHints(ctx)
 		if err != nil {
 			t.Fatalf("HideHints failed: %v", err)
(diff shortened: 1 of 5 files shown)


y3owk1n:
All three valid, all three fixed. Two of them were caused by my change; the third showed my fix did not do what I claimed.

**Issue 1 — accumulating alerts.** Correct, and self-inflicted. Showing the dialog inline had a property I did not notice I was removing: it serialized failed reloads, because a second could not start until the first was dismissed. Off the reply's path, a run of failures stacks a dialog and parks a goroutine behind each. One outstanding alert is now enforced; once dismissed, the next failure alerts again rather than being suppressed for good. A test drives five failing reloads against an undismissed alert, asserts one was shown, then dismisses and asserts the next is shown — it fails against the unguarded version.

**Issue 3 — the deadline was ineffective.** Also correct, and the more important of the two, because it means my first attempt advertised a bound that did not exist. `accessibility.Adapter.Scroll` takes a `context.Context` and discards it outright, and the adapters that do consult one use a pre-flight `checkContext`, so nothing interrupts a call already blocked inside the platform API. A deadline would only ever have helped a caller that returns on its own.

Replaced with a watchdog per test: if the test has not finished within 30 seconds the run ends with a message naming the likely cause, and the stack that comes with it shows which call is stuck. That works whether or not the callee honours a context, which is the point. Verified by forcing the budget to 1ns — the run ends with the diagnosis instead of hanging; restored, the suite passes in 0.3s.

**Issue 2 — shared deadline across subtests.** Correct as written, and it disappears with the above: there is no longer a shared context to misallocate. The tests use `context.Background()` again, which is honest about what these adapters actually respect, and the watchdog is the bound.

The PR body now covers all three, including why the deadline approach was wrong rather than just replacing it quietly.

Gate: `just fmt`, `just lint`, the full `just test` including the integration suite, `just build`, and the race detector over the touched package — all green.

y3owk1n:
Valid, and it is the same finding as the previous round's "deadline spans unrelated subtests" — which I said had dissolved when I dropped the context. It had not. I moved the shared budget from the context to the watchdog and did not notice I had carried the property across. Thanks for re-raising it.

Each subtest now starts its own watchdog, so a test function holding four of them gives each the full budget instead of splitting one between them.

Demonstrated rather than asserted, with the budget temporarily set to 900ms and four subtests sleeping 400ms each — a workload where nothing is hung but the total exceeds the budget:

```
one shared watchdog (previous shape):
  panic: TestScrollServiceIntegration: a real system adapter did not answer within 900ms …
  FAIL

one watchdog per subtest (this change):
  ok    github.com/y3owk1n/neru/internal/app/services   2.097s
```

A genuine hang is still caught: forcing the budget to 1ns ends the run with the diagnosis and a stack, now naming the subtest — `TestHintServiceIntegration/ShowHints_integration: a real system adapter did not answer within 1ns …`.

One implementation note. I first wrapped `t.Run` in a `boundedRun` helper so the watchdog was invisible at the call sites, but `thelper` then reads every subtest body as a helper function missing `t.Helper()` — fifteen lint failures for closures that are test bodies, not helpers. The watchdog is now the first line inside each subtest, which is idiomatic and puts the budget where it applies.

Setup outside a subtest is left to the `go test -timeout` backstop; the call that actually hangs is inside one.

Gate green: `just fmt`, `just lint`, the full `just test` including the integration suite, and `just build`.

PR #1146, opened 2026-08-02
test(integration): skip live-input tests without permission, and stop caching

y3owk1n:
## Description

`just test` was red on a developer machine for reasons no code change could fix, and green in CI. Three separate things were making the suite lie about itself.

**Six tests need macOS Accessibility permission and failed without it.** They drive real input or read the live AX tree, both behind the same TCC gate, and a `go test` binary is not normally granted it. Without the grant the window server drops posted moves and the AX query finds no frontmost window — so the cursor "settles" exactly where it started, `Register` cannot create a CGEventTap, and all six report a failure *of the code*. What failed is the environment:

```
MoveCursorToPoint((687,754)) from (677,744) settled at (677,744)
ClickableElements() error = failed to get frontmost window
Register("cmd+alt+ctrl+shift+f12") error = failed to register hotkey
```

They now skip when the permission is absent, with a message saying so and where to grant it. The suite already skipped on `testing.Short()` and on `"no active screen bounds"`, so this follows an existing pattern rather than inventing one.

**The integration recipes now pass `-count=1`.** Go keys its test cache on the inputs it can see — sources, env, files read — and none of those change when the thing an integration test actually depends on does: whether the permission is granted, whether a daemon holds the socket, whether the screen is locked. A result cached under different conditions gets replayed as a pass, which is worse than no result: it reports green for a run that never happened. It did exactly that here — a stale pass hid the hotkey failure from me for several minutes while I was reporting the suite as green.

**`TestAccessibilityAdapterIntegration` ran its real AX scans on `context.Background()`.** A scan of a slow or unresponsive frontmost window blocks in the Objective-C bridge, so with no deadline that is not a slow test but `just test` never returning, with no indication of which test wedged. It now runs under a 30s budget — far above the daemon's own 5s per-scan ceiling, so it can only trip on a real hang.

## Target Platform

- [x] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

## Cross-Platform Checklist

The new helper is `permission_integration_darwin_test.go`, named to match the `*_integration_<os>_test.go` convention so it falls under the depguard exemption. My first name for it did not, and the linter caught it — worth noting the rule is enforced, not just documented.

## General Checklist

- [x] Tests pass (`just test`) — and `just test-race`, both genuinely uncached

## Additional Context

**The gates were verified in both directions, which is the thing worth checking.** A skip that can never un-skip is not a fix, it is deleting the test. With the probe reading the real permission, all six skip and both packages pass. With the probe forced open, exactly the same six run and fail again with the identical errors. So the skip is driven by the permission and nothing else.

**CI coverage is unaffected.** I checked the macOS job log on #1145 before changing anything: `macos-latest` passes all six for real (`TestAccessibilityAdapterIntegration` and both cursor tests, `--- PASS`). The runner has the grant, so they keep running there on every PR. This PR only changes what happens on a machine that lacks it.

**`MoveCursorToPoint bypassSmooth` is gated although it passes without the permission.** It only asserts that the call returns no error, and that call succeeds while the cursor never moves — so its pass says nothing about the path it exists to cover. Gating it makes the skip list coherent (everything that drives real input) and stops it reading as coverage it does not provide. Happy to drop that one from the PR if you would rather keep the diff to the failing tests.

**Not fixed here, but noticed:** `TestConcurrentMovesKeepCursorInViewport` `continue`s out of its assertion on every round when Accessibility Zoom is not engaged, so outside a zoomed session it asserts nothing and is really a race/panic smoke test. That is a coverage question rather than a flakiness one, so I left it alone.

### DIFF
--- a/internal/core/infra/accessibility/adapter_integration_darwin_test.go
+++ b/internal/core/infra/accessibility/adapter_integration_darwin_test.go
@@ -28,12 +28,15 @@ func TestAccessibilityAdapterIntegration(t *testing.T) {
 	adapter := accessibility.NewAdapter(log, nil, nil, client, false)
 	system := darwinplatform.NewSystemAdapter()
 
-	// Bounded rather than context.Background(): every call below reaches the
-	// real AX API, and a scan of a slow or unresponsive frontmost window blocks
-	// in the Objective-C bridge. With no deadline that is not a slow test, it is
-	// `just test` never returning — and the run has to be killed by hand to find
-	// out which test it was. The budget is far above the daemon's own per-scan
-	// ceiling (modes.HintTimeout, 5s), so it can only trip on a real hang.
+	// Bounded rather than context.Background(), but note what this does and
+	// does not buy: the AX client discards the context and calls straight into
+	// the Objective-C bridge, so the deadline cannot interrupt a query that is
+	// already wedged there. What it does do is stop the scan's per-source
+	// goroutines from starting further work once it has expired.
+	//
+	// The actual hang guard is runWithinBudget, which watches from outside the
+	// call. The budget is shared and sits far above the daemon's own per-scan
+	// ceiling (modes.HintTimeout, 5s), so neither can trip on a merely slow run.
 	ctx, cancel := context.WithTimeout(context.Background(), integrationScanBudget)
 	defer cancel()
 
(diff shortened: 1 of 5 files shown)


y3owk1n:
Greptile's finding is valid — fixed in 20899893.

I checked it against the code rather than taking it at face value, and it holds: `InfraAXClient.FrontmostWindow` declares `_ context.Context` and calls straight into the Objective-C bridge, so nothing observes cancellation once a query is inside it. The `select { case <-ctx.Done(): ... default: }` in `ClickableElements` is a pre-check before `queryFunc()`, not a way out of it. The deadline I added bounded nothing that mattered.

The scan now runs under a watchdog *outside* the call, which fails in 30 seconds and names what wedged. Verified by making a stand-in call block past a shortened budget and confirming the failure fires with the intended message, rather than assuming it would.

I kept the bounded context, with an honest comment about what it does and doesn't buy: the scan's per-source goroutines do check it before starting further work, so it limits how much *new* work begins after expiry — it just cannot interrupt a call already in the bridge.

One correction to my own PR description while I'm here: I wrote that without a deadline `just test` would never return. That overstated it. Go's own `-timeout` still applied, so the real symptom was a panic with a full goroutine dump after the default 10 minutes, with no indication of which call was stuck — bad, but bounded. The watchdog turns that into a 30-second failure that says what happened.

`just test`, `just test-race`, `just lint` and `just build` are all green locally, uncached.

PR #1179, opened 2026-08-04
feat(smooth_cursor): animate relative cursor moves with a fixed per-move duration

sboesebeck:
## Description

This PR makes smooth-cursor animation cover relative (keyboard-driven) mouse movement — the hjkl-style `move_mouse_relative` steps that previously always warped instantly even with `[smooth_cursor]` enabled. With `move_mouse_enabled = true`, all cursor movement now animates.

The motivation is visual ergonomics: a cursor that teleports in discrete steps under held-key repeat is hard to follow with the eyes; a short glide keeps it trackable and is noticeably less straining. Relative moves deliberately do not reuse the jump animation's distance-proportional duration: that math produces constant velocity, so a 50px step would move no faster than a 10px one. Instead each relative move animates over a fixed duration (`relative_movement_duration`, default 50ms), so cursor speed scales with the per-move delta — and composes naturally with `held_repeat` acceleration (#1120), whose growing deltas translate directly into growing speed. While an animation is in flight, the next delta extends its endpoint rather than restarting from the current position, so no part of a delta is lost under key repeat.

macOS only for now; on Linux and Windows relative moves stay instant (documented in the platform tables). With smooth cursor disabled (the default), behavior is completely unchanged.

## Related Issues

None. (CONTRIBUTING suggests an issue first for non-trivial changes — happy to open one for discussion and link it here if you prefer that flow.)

## Target Platform

- [x] Platform-agnostic (shared logic, no OS-specific code) — config surface only
- [x] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [x] `feat` — New feature

## Cross-Platform Checklist

- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no new port method: the darwin adapter opts into the existing optional relative-move extension, and where it does not apply callers take their existing instant-warp fallback, which is exactly the previous behavior.

## General Checklist

## Config Changes

One new key under `[smooth_cursor]`; existing configs keep working unchanged. No new toggle — `move_mouse_enabled` now covers relative movement too (see behavior note below).

| Key | Type | Default | Meaning |
| --- | ---- | ------- | ------- |
| `smooth_cursor.relative_movement_duration` | int | `50` | Fixed animation duration per relative move in ms (>= 10), macOS only |

```toml
[smooth_cursor]
move_mouse_enabled = true
relative_movement_duration = 50   # matches the held_repeat interval so animations chain seamlessly
```

The key is settable at runtime via `neru config set` and hot-reloads like the rest of `[smooth_cursor]`.

To see it live: enable the snippet above, then hold a `move_mouse_relative` binding (default hjkl in pointer contexts) or run `neru action move_mouse_relative --dx=50 --dy=0`.

## Behavior Change for Existing Smooth-Cursor Users

Users who already run with `move_mouse_enabled = true` will see relative (hjkl) moves start animating after this change — previously they always warped. This is the intended reading of the option (smooth cursor covers cursor movement, full stop) and the default 50ms glide is subtle; there is currently no separate opt-out for relative moves while keeping jump animation. If that combination turns out to be wanted, an opt-out can be added later.

## Additional Context

Two alternatives were considered and rejected:

- **Distance thresholds** (animate small/large moves differently based on a configurable cutoff): the dispatcher already knows whether a move is a jump or a relative step, so a threshold would only reconstruct — imprecisely — information the code has. The semantic split keeps both animation models exact and needs no user calibration.
- **Reusing the jump animator as-is for relative moves**: rejected for the constant-velocity math above, and because restarting each repeat from the mid-animation cursor position silently discards the un-animated remainder of every delta, capping effective speed at `1 / duration_per_pixel` regardless of the configured step size.

The implementation is minimally invasive: no port signature changes, no behavior change unless smooth cursor is enabled. Endpoint extension is done atomically under the animator's existing lock so two concurrent held-repeat sources (mode repeat and hotkey repeat) compose their deltas instead of losing one — pinned by a new `-race` test with 8 concurrent movers. The pending endpoint is clamped to the active screen so holding a key at a screen edge cannot make the target drift off-screen and delay the way back.

Review update: the first revision introduced a separate `[smooth_cursor.relative]` section with its own `enabled` toggle; per review feedback this was folded under `move_mouse_enabled`, and the remaining single tuning key was flattened to `relative_movement_duration`.

### DIFF
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -392,6 +392,12 @@ steps = 10                                  # Number of intermediate positions
 max_duration = 200                          # Maximum animation duration in ms
 duration_per_pixel = 0.1                    # Ms per pixel for adaptive duration
 
+# Smooth cursor for relative (keyboard-driven) movement, e.g. move_mouse_relative.
+# Uses a fixed duration per move so larger deltas move proportionally faster.
+[smooth_cursor.relative]
+enabled = false                             # Animate relative mouse movement (macOS only)
+duration = 50                               # Fixed animation duration per move in ms
+
 # Smooth scroll (animates scroll actions with eased chunking)
 # See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#smooth_scroll
 [smooth_scroll]
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -391,12 +391,7 @@ move_mouse_enabled = false                  # Enable smooth mouse movement
 steps = 10                                  # Number of intermediate positions
 max_duration = 200                          # Maximum animation duration in ms
 duration_per_pixel = 0.1                    # Ms per pixel for adaptive duration
-
-# Smooth cursor for relative (keyboard-driven) movement, e.g. move_mouse_relative.
-# Uses a fixed duration per move so larger deltas move proportionally faster.
-[smooth_cursor.relative]
-enabled = false                             # Animate relative mouse movement (macOS only)
-duration = 50                               # Fixed animation duration per move in ms
+relative_duration = 50                      # Fixed duration per relative (hjkl) move in ms (macOS only)
 
 # Smooth scroll (animates scroll actions with eased chunking)
 # See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#smooth_scroll
--- a/configs/default-config.toml
+++ b/configs/default-config.toml
@@ -391,7 +391,7 @@ move_mouse_enabled = false                  # Enable smooth mouse movement
 steps = 10                                  # Number of intermediate positions
 max_duration = 200                          # Maximum animation duration in ms
 duration_per_pixel = 0.1                    # Ms per pixel for adaptive duration
-relative_duration = 50                      # Fixed duration per relative (hjkl) move in ms (macOS only)
+relative_movement_duration = 50             # Fixed duration per relative (hjkl) move in ms (macOS only)
 
 # Smooth scroll (animates scroll actions with eased chunking)
 # See https://github.com/y3owk1n/neru/blob/main/docs/CONFIGURATION.md#smooth_scroll
--- a/internal/adapter/platform/darwin/mouse_animator.go
+++ b/internal/adapter/platform/darwin/mouse_animator.go
@@ -103,6 +103,54 @@ func (a *smoothCursorAnimator) pendingTarget() (image.Point, bool) {
 	return a.pendingEnd, a.hasPending
 }
 
+// settle finishes any in-flight animation immediately: the worker is canceled
+// and the cursor warps straight to the endpoint it was animating toward.
+// Position-dependent reads call this so an action firing mid-animation acts
+// at the point the user aimed for, without waiting the animation out.
+func (a *smoothCursorAnimator) settle() {
+	end, ok := a.takePendingForSettle()
+	if !ok {
+		return
+	}
+
+	eventType, button := dragEventType()
+	pos := C.CGPoint{x: C.double(end.X), y: C.double(end.Y)}
+	C.NeruMoveMouseWithTypeAndButton(pos, eventType, button)
+}
+
+// takePendingForSettle atomically ends the in-flight animation and hands back
+// its endpoint: the worker is canceled with the same generation bump as
+// stop() — so a step already past its checks is dropped rather than landing
+// after the settle warp — waiters are released, and the pending endpoint is
+// cleared. ok reports whether there was an animation to settle.
+func (a *smoothCursorAnimator) takePendingForSettle() (image.Point, bool) {
+	a.mu.Lock()
+
+	if !a.hasPending {
+		a.mu.Unlock()
+
+		return image.Point{}, false
+	}
+
+	end := a.pendingEnd
+	stopCh := a.stopCh
+	done := a.done
+	a.reqCh = nil
+	a.stopCh = nil
+	a.done = nil
+	a.generation++
+	a.hasPending = false
+
+	if stopCh != nil {
+		close(stopCh)
+	}
+	a.mu.Unlock()
+
+	done.close()
+
+	return end, true
+}
+
 // currentGeneration returns the animation generation in effect right now.
 func (a *smoothCursorAnimator) currentGeneration() uint64 {
 	a.mu.Lock()
--- a/internal/adapter/platform/darwin/mouse_animator.go
+++ b/internal/adapter/platform/darwin/mouse_animator.go
@@ -46,6 +46,7 @@ type cursorRequest struct {
 	button           uint32
 	maxDuration      int
 	durationPerPixel float64
+	fixedDuration    int // when > 0, overrides the distance-derived duration
 	done             *cursorAnimationDone
 }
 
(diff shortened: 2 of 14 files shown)


y3owk1n:
Thanks for the PR, looks good, i am just thinking do we really need a new `smooth_cursor.relative`, or we just rely on the `smooth_cursor.enabled`. Since this looks like a missing implementation that suppose to work with any cursor movement if enabled. What do you think?

sboesebeck:
Agreed — that's the better reading: relative movement animating is the missing piece of smooth cursor, not a separate feature. Reworked in the latest commit: `move_mouse_enabled` now governs relative moves too, and since that left the `[smooth_cursor.relative]` table with a single key, I flattened it to `smooth_cursor.relative_duration` (default 50ms) beside `max_duration` / `duration_per_pixel`.

One heads-up, also noted in the PR body: users already running with `move_mouse_enabled = true` will see hjkl moves start animating after this — previously they always warped. The default 50ms glide is subtle, and if anyone ever wants jump animation without relative animation, an opt-out can be added later. Docs, example config, validator, and tests are updated; `just ci` is green.

internal/config/validators_pointer.go
```diff
@@ -52,6 +52,13 @@ func (c *Config) ValidateSmoothCursor() error {
 		)
 	}
 
+	if c.SmoothCursor.RelativeDuration < 1 {
+		return derrors.New(
+			derrors.CodeInvalidConfig,
+			"smooth_cursor.relative_duration must be >= 1",
```
y3owk1n: in `mouse_animator.go`, we have set a minimum for `minAnimationDuration` as 10, so this >= 1 is technically not true, as if it's lesser than 10, it will just use the minimum that we hardcoded previously. Since we are doing a default of 50, maybe we should validate this as >= 10? What do you think?

internal/adapter/platform/darwin/mouse_animator.go
```diff
@@ -151,37 +167,77 @@ func (a *smoothCursorAnimator) animateTo(end image.Point, steps int, eventType,
...
+	eventType, button uint32,
+) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	base := a.pendingEnd
+	if !a.hasPending {
+		base = CursorPosition()
 	}
+
```
y3owk1n, line 246: if the new end is the same as pending end, we are still submitting. I think we could have an early return guard here if we dont have to animate at all.

internal/config/validators_pointer.go
sboesebeck: Good catch — done. It now validates >= 10 via a `MinSmoothCursorRelativeDuration` constant next to the other config defaults, cross-referenced with the animator's `minAnimationDuration` in comments on both sides so the two can't drift apart unnoticed. Docs table updated to match.

internal/adapter/platform/darwin/mouse_animator.go
sboesebeck: Agreed — added an early return when the clamped endpoint equals the pending one (zero delta, or a delta the screen-edge clamp ate). Without it the no-op submit would cancel and restart the animation already heading there. Pinned by a new test asserting no-op moves leave the queued animation untouched.

internal/app/services/action_service.go
y3owk1n, line 245: this returns once the animation is submitted, not when it settles. Unlike the jump paths, nothing here waits for cursor idle and all the potential actions resolves their point from `CursorPosition()`, so if any of them fires within the duration, it reads a the cursor position in the middle and performa those action at there while still animating to the end point.

i don't think we should just wait for idle here, that adds up to 50ms of latency to every click after a move. Since the animator already tracks the pending endpoint, maybe we could have a `settle()` that stops the animation and warps straight to `pendingTarget()` (the generation bump in `stop()` should already make that safe), and call it before position-dependent reads. What do you say?

configs/default-config.toml
```diff
@@ -391,6 +391,7 @@ move_mouse_enabled = false                  # Enable smooth mouse movement
 steps = 10                                  # Number of intermediate positions
 max_duration = 200                          # Maximum animation duration in ms
 duration_per_pixel = 0.1                    # Ms per pixel for adaptive duration
+relative_duration = 50                      # Fixed duration per relative (hjkl) move in ms (macOS only)
```
y3owk1n: i am just thinking if we should make the naming of this config more verbose? Maybe `relative_movement_duration` or some sort. It feels like hard to understand what it does in a quick glance.

internal/app/services/action_service.go
sboesebeck: You're right, and the settle() approach is the right call — implemented as suggested. Since every position-dependent read (click resolution, save_cursor_pos, the indicators) funnels through the one cursor-position read on the system adapter, settle() hooks in exactly there: it takes the pending endpoint under the animator lock with the same generation bump as stop() (so a step already past its checks is dropped rather than landing after the warp), releases waiters, and warps straight to the endpoint. A read mid-animation now returns the point the user aimed for, with no added latency when nothing is animating. Pinned by a test covering the state transition: endpoint handed back exactly once, waiters released, stale steps refused.

internal/config/config_defaults.go
```diff
@@ -121,6 +121,16 @@ const (
...
 
+	// DefaultSmoothCursorRelativeDuration is the default fixed animation
+	// duration for relative moves (ms). Matches DefaultHeldRepeatInterval so
+	// animations chain seamlessly under held-key repeat.
+	DefaultSmoothCursorRelativeDuration = 50
+
+	// MinSmoothCursorRelativeDuration is the smallest accepted
+	// relative_duration (ms). Mirrors the animator's minAnimationDuration
+	// floor (darwin adapter): anything below it would be silently rounded up.
+	MinSmoothCursorRelativeDuration = 10
```
y3owk1n: seems like this is only mainly being used for config validation and tests. If this should always tally with `minAnimationDuration`, i think we should just use it instead of redeclaring the same value both place, and needs to keep in mind to ensure both are updated when one of them changes.

configs/default-config.toml
sboesebeck: Agreed, renamed to `relative_movement_duration` as suggested — struct field, validator message, example config, and both doc locations updated.

internal/config/config_defaults.go
sboesebeck: Done — collapsed into one constant. Since the One Rule forbids config importing the darwin adapter, the shared constant lives in config as `MinSmoothCursorAnimationDuration` and the animator's hardcoded `minAnimationDuration` is gone: both the animation floor and the validator bound read the same declaration now, so they can't drift apart.

internal/app/services/action_service.go
sboesebeck: Correction to my earlier reply: hooking settle() into the cursor-position read itself turned out to be too aggressive — the mode-indicator follower polls the position while a mode is active, and every poll settled the animation, so relative moves stopped animating entirely with an indicator on screen. Fixed in the latest commit: the position read is a pure observation again, and settling is an optional `CursorSettler` port extension (same pattern as `CursorSynchronizer`) that only the action paths call when resolving their target point from the cursor — clicks and `save_cursor_pos`. Mid-animation actions still land on the aimed-for point; observers no longer cut animations short.

y3owk1n:
Amazing work and thanks for all the quick turn around. I am pretty happy with the current implementation. Let's wait for the CI to turn all green, and we are good to merge.

y3owk1n (approved):

y3owk1n:
Thanks!

Issue #1211, opened 2026-08-05
Grid and recursive grid draw through Frames


y3owk1n:
## Parent

#1203

## What to build

Grid and recursive grid draw through **Frames**.

Both surfaces convert to handing over a Frame on entry and exit. Per ADR 0003, per-keystroke updates — the grid's match prefix and its hide-unmatched flag — keep their direct call path and are **not** routed through frame construction. Grid narrowing must feel exactly as it does today; `AGENTS.md` is unambiguous that key handling getting slower is a regression.

Measure it rather than assume it. If the transition path has moved the per-keystroke cost at all, say so on the ticket before merging.

## Acceptance criteria

- [ ] Entering and leaving grid and recursive grid both go through Frames
- [ ] Per-keystroke grid updates still use direct calls; no frame is constructed on a keystroke
- [ ] The recursive-grid draw no longer takes ten positional parameters from the app layer
- [ ] Per-keystroke cost in grid mode is measured and reported as unchanged
- [ ] Existing grid and recursive-grid simulation journeys pass

## Blocked by

- #1210

y3owk1n:
Per-keystroke cost measured before merging, as the ticket asks.

Two benchmarks drive the real application through the simulation harness
(`internal/app/simulation_benchmark_test.go`), one keystroke per iteration,
with the keystroke that resets the input untimed. Runs of this branch and
`main` interleaved to control for drift, compared with `benchstat` (n=8,
Apple M3):

| | time | allocations |
| --- | --- | --- |
| grid narrowing keystroke | ~11.1µs → ~11.3µs, p=0.57 | 74 → 74, identical bytes |
| recursive-grid keystroke | ~4.50µs → ~4.57µs, p=0.28 | 63 → 64, +128 B (p=0.000) |

**Grid narrowing is unchanged**, in time and allocations both — the acceptance
criterion. It still goes through the direct calls: no frame is constructed on a
narrowing keystroke.

**The transition path did move the recursive-grid keystroke, by one
allocation.** That mode has no incremental path — every key repaints its whole
surface — so its per-keystroke update *is* the frame, and boxing the frame into
the port's interface costs 128 bytes per key. There is no measurable time
difference above noise. Saying so here rather than leaving it implied, since
that is the cost ADR 0003 predicted a declarative port would charge on a hot
path; it is paid only where the surface was already being repainted anyway.

PR: #1221

Issue #1210, opened 2026-08-05
Frame port, with hints as the first surface


y3owk1n:
## Parent

#1203

## What to build

Introduce the **Frame** — the complete description of what should be on screen for one mode — and convert hints to it.

Showing an overlay is currently a three-call ritual: show the window, switch it to the mode, then draw. The order is not expressed anywhere and nothing checks it; thirteen call sites are each trusted to get it right. Getting it wrong shows an empty overlay or leaves the previous mode's on screen.

A mode hands over a Frame; the adapter owns the sequence needed to realise it. A Frame carries only domain values — never a Style, a render model, or a platform handle. That constraint is what later makes the layering ban expressible, so it is not negotiable.

Hints is the first surface through the new path, end to end.

## Acceptance criteria

- [ ] The port gains a Frame-carrying transition method; the show/switch/draw sequence is owned in one place
- [ ] A Frame carries domain values only — no Style, render model, or platform type
- [ ] Entering and leaving hints goes through it; the hints path issues no direct manager calls
- [ ] Existing hints simulation journeys pass, asserting on domain values recorded from Frames
- [ ] Tests asserting a specific manager method was called in a specific order are deleted, not ported

## Blocked by

- #1207 — a Frame carries domain values only, which requires the overlay to already own Style.

Issue #1207, opened 2026-08-05
The overlay resolves its own Style


y3owk1n:
## Parent

#1203

## What to build

The overlay resolves its own **Style** — an overlay's appearance, being its configuration combined with the current light/dark theme.

That derivation happens in five places across three layers today. When the system appearance changes, the app has to poke six render components, then the renderer, then re-dispatch a per-mode refresh; a missed one leaves an overlay drawn in the previous theme's colours until the mode is re-entered.

Move ownership to the adapter, which already holds the theme provider and gains the configuration. Resolve on config or theme change, **not per draw**, so the draw path allocates no more than it does today.

Delivers: a theme or config change reaches every overlay through one notification instead of a seven-way fan-out.

## Acceptance criteria

- [ ] Style is built in one place; no other layer derives it
- [ ] The app's theme-change handler no longer fans out to render components individually
- [ ] A config reload and a theme change both reach an overlay that is currently on screen
- [ ] A test pins that Style is re-resolved once per config or theme change, not once per draw
- [ ] A simulation journey asserts a theme change reaches a visible overlay

## Blocked by

None — can start immediately.

Issue #1205, opened 2026-08-05
Headless is a declared capability, not a null window handle


y3owk1n:
## Parent

#1203

## What to build

*Prefactor for #1203's construction move.* A headless environment — no display server — is currently detected by inferring it from a null native window handle. Replace that inference with a capability the overlay declares explicitly.

Behaviour is unchanged: the same environments are headless afterwards. What changes is that headlessness is stated rather than guessed, and the simulation harness stops depending on a null handle to reach its headless path.

This lands before the window handle is removed, so that the later change is a deletion rather than a redesign under pressure.

## Acceptance criteria

- [ ] Headless is a declared capability on the overlay, reached the same way other optional capabilities are — by type assertion, not by widening the port
- [ ] The component factory's headless path keys off the capability, not the window handle
- [ ] The simulation harness declares itself headless explicitly; its recorder no longer relies on a null handle
- [ ] Existing simulation journeys pass unchanged
- [ ] A test pins that an overlay which cannot render reports itself headless

## Blocked by

None — can start immediately.

PR #1219, opened 2026-08-06
refactor(overlay): let the overlay build the render components it draws through

### LINKED ISSUE #1209: The overlay constructs its own render components
## Parent

#1203

## What to build

The overlay constructs and owns its own render components. The app hands it configuration and a theme provider, and nothing else.

Today the app builds six render components and injects them back through six setter methods matched by six getters — the single largest block of the overlay interface, and the reason the app layer names a platform window-handle type at all.

Delivers: twelve methods and the platform handle leave the app layer. Overlay construction becomes one startup phase inside the adapter rather than several in the app.

The startup phase sequence is numbered and individually unwound; it must keep unwinding exactly as it does today. This change is more likely to break shutdown than startup — check the unwind path explicitly.

## Acceptance criteria

- [ ] The overlay builds its own render components from configuration and a theme provider
- [ ] The six setters, six getters, and the window-handle accessor are gone from the interface
- [ ] No app-layer package names a platform pointer type
- [ ] Startup still unwinds cleanly from a failure in every phase it could fail in before
- [ ] Existing simulation journeys pass; the harness builds through the capability from its blocking ticket

## Blocked by

- #1205 — headless must have an explicit signal before the window handle is removed.
- #1207 — the overlay cannot sensibly construct components whose Style it does not yet own.

y3owk1n:
## Description

This PR moves construction of the overlay's render components out of the
application layer and into the overlay itself. The app used to build all six
of them, then hand each one back through a setter, and read them back through
a matching getter — twelve methods on the overlay's contract that existed only
because the wrong layer was doing the building. It is also the only reason the
app layer ever had to name the platform's native window handle.

The overlay is now handed the configuration and a theme provider, and builds
what it draws on the surface only it knows about. A configuration reload or a
theme change reaches those components through one notification instead of six,
and each platform backend states for itself whether it has a surface to build
against at all, rather than that being inferred elsewhere.

Nothing changes about what appears on screen, how it looks, or how fast a
keystroke is handled. Startup still runs the same numbered phases in the same
order, with the same failure still failing the same phase.

## Related Issues

Closes #1209. Part of #1203 (step 4 of its sequencing), and follows #1205 and
#1207, which it was blocked on. It does not attempt the Frame port steps
(#1210, #1211) — the handful of app-layer call sites that still talk to a
render component directly are left for those.

## Target Platform

- [x] macOS
- [x] Linux
- [x] Windows

## Type of Change

## Cross-Platform Checklist

- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no new capability is being added; every backend implements the two new contract methods outright, and the headless one builds nothing rather than reporting a failure.

## General Checklist

- [x] Documentation updated (if applicable) — the overlay area contract, the
      startup phase list in the architecture doc, and the ADR this sequence
      records itself in. No user-facing config or command surface changed, so
      the configuration and CLI references are untouched.

## Additional Context

**No config, command, flag or environment surface changes.** Nothing a user
writes or types is affected.

The unwind path got the explicit check the issue asked for. The phases, their
numbering and their cleanup closures are unchanged; the only failure that
could fail the UI-components phase before — the virtual pointer overlay
failing to build — still fails it, and every other component still degrades to
a warning and a mode that does not draw. One thing is now better than it was:
when that failure happens, the components already built are handed to the
overlay before the error goes back, so a later teardown reaches the native
windows they hold instead of orphaning them.

Each backend now decides for itself whether it has a surface, and there are
tests on macOS and Linux pinning that the build believes that declaration
rather than re-deriving it from a handle it can see — which is what the app's
deleted headless check used to guarantee. The one call that reaches the
overlay's shared state straight through the contract carries the nil-receiver
guard the rest of this package documents, since a backend that finds no
display server is handed out as a typed nil.

Two deliberate limitations. The build assertions covering every component run
off macOS only, because there the three indicators each allocate a real
window, which is not something a unit test should do; what they are built from
is shared code, so the wiring under test is the wiring macOS runs. And the
three indicators still resolve their own colours at draw time from the theme
provider they are now handed at construction — folding those into the single
style owner was not part of this step, and the area contract says so.

`just lint-cross` was run as well as `just ci`, since the Linux and Windows
backends are build-tagged and the local gate cannot see them.


y3owk1n:
Fixed in 759a16e3 — the UI-components phase now releases the overlay's windows when a component build fails partway. The phase's cleanup closure is registered only after the phase succeeds, so the failing call had to do it itself. (The leak predates this PR: the app previously held the same partial components in its own fields and dropped them just as silently.)

PR #1244, opened 2026-08-06
test(app): cover a display change that lands while a mode is closed

### LINKED ISSUE #1233: Refresh a mode's overlay on activation when the display changed while it was inactive
Follow-up to #1225, which deleted three write-only `*OverlayNeedsRefresh` flags. Those flags were set whenever the display configuration changed while a given mode was inactive, and cleared when that mode next activated. Nothing ever read them, so the behaviour they were reaching for was never delivered. #1225 deleted the flags rather than carrying them forward; this issue states the behaviour they hinted at, so it can be judged as a feature on its own terms instead of being inferred from dead state.

Parent spec: #1224 ("Out of Scope": *"Lazy refresh on activation. Deleting the write-only flags does not implement the behaviour they hinted at. If it is wanted, it is a feature with its own stated behaviour."*)

### Category

UI / Overlays

### Target platform

All platforms

### Problem or use case

The implied need: when a display is plugged in, unplugged, or reconfigured while a mode is *inactive*, that mode should not come back sized or positioned for a screen that no longer exists. The active mode is already handled — `processScreenChange` in `internal/app/lifecycle.go` delegates to `RefreshGridForScreenChange` / `RefreshHintsForScreenChange` / `RefreshRecursiveGridForScreenChange`. The inactive modes were what the flags were tracking.

**However, an audit while doing #1225 suggests the correctness gap may already be closed**, and this issue should be judged with that in mind:

- Every mode activation rebuilds against current screen bounds unconditionally. `createGridInstance` (`internal/app/modes/grid.go`) calls `system.ScreenBounds` on every activation; recursive grid does the same in its activation path; hints regenerates its whole collection from the accessibility tree.
- Every activation that brings up the shared overlay window runs `ResizeToActiveScreen` inside `Adapter.ShowFrame` (`internal/adapter/overlay/adapter.go`), so the window is sized to the current active screen regardless of what happened while the mode was inactive.

So the eager path appears to already deliver the outcome the flags were aiming at, which is consistent with nothing having read them and no bug report existing.

That leaves two things genuinely worth deciding:

1. **Verification.** There is no test covering "display changes while mode X is inactive, then mode X is activated". #1226 adds screen-change journeys for the *active* mode only. If the eager path is the mechanism we rely on, that reliance should be pinned by a journey rather than assumed.
2. **The inverse optimisation.** If a rebuild-on-every-activation is more work than needed, the interesting version of "lazy" is the opposite of what the flags implied: track that the display *has not* changed and skip work on activation. Whether that is worth it is a latency question, and `AGENTS.md` says latency is the product — but it needs a measurement first, not a flag.

### Proposed solution

Decide between three outcomes, in this order:

1. **Close as already-satisfied**, after adding a simulation journey per mode that drives: activate mode X, exit, change screen parameters via the app's screen-parameters-changed entry point (added by #1226), re-activate mode X, and assert the Frame reaching the fake overlay is built for the new screen. This is the cheapest outcome and the most likely correct one.
2. **Implement lazy refresh** only if journey (1) shows a mode that genuinely comes back stale. In that case the stated behaviour is: on activation, if the display configuration changed since this mode last drew, rebuild from current screen bounds before drawing; otherwise proceed as today. The trigger must be a comparison of actual display geometry, not a boolean set by an observer — a boolean cannot distinguish "changed and changed back".
3. **Implement the skip optimisation** only if a measurement shows the unconditional rebuild costs meaningful activation latency.

Whichever is chosen, it should land after #1224's mode-extension work, so that any per-mode refresh behaviour is expressed as a mode extension rather than as a fourth near-identical function in the app layer.

### Alternatives considered

- **Restore the deleted flags.** Rejected in #1225 and rejected here: a boolean set by an observer and cleared on activation cannot express "the display changed and changed back", and it records a fact the activation path can simply ask for directly.
- **Do nothing and leave it undocumented.** That is how the flags survived this long. Filing it means the next contributor reading the screen-change path finds a decision rather than a puzzle.

### Contribution

No PR planned; filing so the behaviour is recorded rather than inferred from deleted state.

y3owk1n:
## Description

This PR adds test coverage for a display configuration change that lands while a
mode is closed — a monitor plugged in, unplugged or set to a different
resolution between two activations — and asserts that the mode comes back built
for the display that exists now rather than the one it last drew on.

Existing coverage only had the mode open when the arrangement changed. The
closed case was the outcome the overlay-staleness flags deleted in #1234 were
reaching for, and nothing pinned it. It turns out to need no change: every
activation already reads the current arrangement unconditionally, which is
consistent with nothing ever having read those flags and no bug report existing.
The behaviour was already correct; only the guarantee was missing.

Four journeys, one per mode with a surface a journey can read — grid and
recursive grid come back over the new display, hints come back regenerated from
the accessibility tree and drawn against it, and the monitor picker comes back
with panels for the arrangement as it now is. Scroll draws no surface of its
own, so there is no frame built for a display to assert on; the reason is
recorded next to the four.

## Related Issues

Closes #1233

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no OS-specific files touched
- [x] No darwin imports from untagged (shared) code — [The One Rule](https://github.com/y3owk1n/neru/blob/main/docs/ARCHITECTURE.md#the-one-rule) — N/A, test-only change in shared code

## General Checklist

- [x] Documentation updated (if applicable) — N/A, no user-facing behaviour, config or command surface changed

## Additional Context

**The journeys were checked to have teeth.** A test that passes the moment it is
written proves nothing on its own, so each of the four was run against a
temporary mutation of its mode that reuses the bounds or targets it last drew
with instead of reading the current arrangement. All four fail under that
mutation, naming the display that is gone; all four pass without it. The
mutations were reverted and are not part of this diff.

**The two alternatives in the issue were not taken.** Lazy refresh on activation
(rebuild only when the display changed since the mode last drew) has nothing
left to fix — the journeys show no mode coming back stale. The inverse
optimisation, skipping the rebuild when the display has not changed, was not
attempted: knowing it has not changed means asking the platform for the current
arrangement, which is the query the skip would be avoiding, so the only saving
available is on the rebuild that follows. That is a measurement the simulation
harness cannot make honestly — its screen queries are in-process fakes, so a
benchmark through it would understate exactly the cost being questioned. Left
for a real measurement if activation latency ever shows up as a problem.

### DIFF
--- a/internal/app/simulation_journey_test.go
+++ b/internal/app/simulation_journey_test.go
@@ -1748,6 +1748,259 @@ func TestSimulation_ScreenChangeWithNothingToDrawExitsTheMode(t *testing.T) {
 	sim.waitMode(domain.ModeGrid)
 }
 
+// The display-change journeys above all have the mode open when the
+// arrangement changes. The four below are the other half of that set: the
+// change lands while the mode is *inactive*, and the mode has to come back
+// built for the display that exists now rather than the one it last drew on.
+//
+// Nothing records that a display changed while a mode was away — each
+// activation reads the current arrangement unconditionally — so what these
+// pin is that eager rebuild. It is the whole mechanism protecting a user who
+// docks, undocks or changes resolution between two activations, and without a
+// journey on it, replacing it with anything cheaper would look free.
+//
+// Four rather than five: scroll draws no surface of its own, so it has no
+// frame built for a display for a journey to read. Its overlay content is the
+// indicators, which are placed by the adapter rather than described in a
+// Frame.
+
+// TestSimulation_GridComesBackOnTheDisplayThatChangedWhileItWasInactive covers
+// the commonest shape of this: a user leaves grid mode, plugs a display in or
+// changes resolution, and comes back to grid.
+//
+// A grid built for the display that is gone is worse than no grid: its cells
+// still have labels, so typing one lands the cursor at a position derived from
+// a screen the user is no longer looking at.
+func TestSimulation_GridComesBackOnTheDisplayThatChangedWhileItWasInactive(t *testing.T) {
+	sim := newSimHarness(t, simConfig(), nil)
+
+	sim.pressHotkey(gridHotkey)
+	sim.waitMode(domain.ModeGrid)
+	sim.waitFor("grid drawn", func() bool { return sim.overlay.lastGrid() != nil })
+
+	if got, want := sim.overlay.lastGrid().Bounds(), localBounds(simScreen); got != want {
+		t.Fatalf("grid drawn over %v before the display change, want %v", got, want)
+	}
+
+	sim.press("Escape")
+	sim.waitMode(domain.ModeIdle)
+
+	drawsBefore := sim.overlay.gridDrawCount()
+
+	sim.changeScreen(simDisplayResized())
+
+	// The mode is not on screen, so the change itself must draw nothing —
+	// that half is TestSimulation_ScreenChangeWhileIdleLeavesTheOverlayHidden;
+	// what matters here is that the rebuild happens on activation instead.
+	if got := sim.overlay.gridDrawCount(); got != drawsBefore {
+		t.Errorf("the grid was drawn %d times by a display change it was not open for",
+			got-drawsBefore)
+	}
+
+	sim.pressHotkey(gridHotkey)
+	sim.waitMode(domain.ModeGrid)
+	sim.waitFor("grid drawn again", func() bool {
+		return sim.overlay.gridDrawCount() > drawsBefore
+	})
+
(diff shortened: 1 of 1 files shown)


y3owk1n:
Valid — fixed in 7b9480cc.

The panel assertions counted two and then required a panel for the display that changed, but never required the unchanged one, so a picker coming back with the changed display twice passed. Both displays are now named individually. Checked against a mutation that duplicates one monitor and drops the other: it passed before, it fails now.

Issue #1232, opened 2026-08-06
Screen-change refresh belongs to the mode


y3owk1n:
## Parent

#1224

## What to build

The last place mode knowledge lives outside the package that owns modes, and the messiest.

A display-configuration change is handled by three near-identical app-layer functions, one per mode. Each checks whether its mode's feature is enabled in configuration, checks whether its mode is the active one, calls that mode's refresh, and returns whether the overlay ended up resized. The caller uses those three results to decide whether a final overlay resize still needs to run. Adding a mode means adding a fourth copy.

Make screen-change refresh an optional extension carried by hints, grid and recursive grid, and collapse the three functions into one dispatch. After this, no switch on the active mode remains outside the modes package.

Three things this ticket must not lose, and they are why it is last:

**The per-mode configuration gate.** A mode whose feature is disabled must not be refreshed. That check has to survive the collapse rather than being dropped as incidental.

**The "was the overlay resized" result.** It decides whether a final resize runs, and getting it wrong either skips a resize the overlay needed or performs one while idle — which would show an overlay the user never asked for.

**Idle.** A screen change while no mode is active must leave the overlay hidden. The current code is explicit about this and the reason is user-visible: resizing the overlay brings it up.

**The lock moves up, per ADR 0004** — the same established pattern as #1229 and #1227's successor, not a fresh decision. Today each of the three refreshers takes the mutex itself and re-reads the active mode as a TOCTOU guard, because the caller snapshots it unlocked. One dispatch under one lock hold removes both the snapshot and the guards.

**#1226 is what makes this safe.** It characterises today's behaviour across all three modes plus idle plus the nothing-to-draw path. If any journey it added has to be edited to make this ticket pass, that is a defect in this change, not in the journey.

#1225 is a prefactor for this ticket: it removes the write-only staleness flags from these same three functions, so what is left to collapse is only the logic that matters.

## Acceptance criteria

- [ ] Screen-change refresh is an optional extension implemented by hints, grid and recursive grid
- [ ] The three app-layer per-mode screen-change functions are collapsed into one dispatch
- [ ] No switch on the active mode remains anywhere outside the modes package
- [ ] A mode whose feature is disabled in configuration is not refreshed
- [ ] The final-resize decision behaves as it does today
- [ ] A screen change while idle leaves the overlay hidden
- [ ] The dispatch takes the handler mutex once; the unlocked snapshot and the per-mode TOCTOU re-checks are gone
[...]

PR #1248, opened 2026-08-06
fix(hints): stop a hung app blocking the keyboard when a display changes

### LINKED ISSUE #1243: Hint generation on the screen-change path runs under the handler lock with no deadline
Follow-up from the #1224 campaign. Anticipated by ADR 0004 and deliberately left open by #1232.

## Neru version

`dev` — source build of `main` at `06414b30`. The condition predates the campaign; what changed is that ADR 0004 now names it and the equivalent path on monitor move has been fixed, leaving this one inconsistent.

## Operating system

macOS

## OS version

macOS 26.5.2 (build 25F84), arm64. The code path is platform-neutral — hint generation goes through the accessibility port, so AT-SPI and UIA backends reach the same unbounded call.

## Navigation mode

Hints

## What happened

On the screen-change refresh path, hint generation walks the accessibility tree **while `h.mu` is held, with no deadline**. `h.mu` serializes every keystroke, so an accessibility tree that stops responding — an unresponsive or hung application in the frontmost position — pins the mode handler and with it all key handling, for as long as the platform call takes to return.

The activation path does not have this problem: it wraps hint generation in the configured `HintTimeout` (5s by default). And #1238 closed the identical gap on the **monitor-move** path, after `deadlock-reviewer` found that moving that walk under one lock hold left it unbounded because a `move_monitor` context carries no deadline. The screen-change path was not given the same bound.

## What did you expect

Every accessibility walk that runs under `h.mu` is bounded by `HintTimeout`, the same as the activation path and the same as monitor move since #1238. A hung tree should cost hints mode, not the keyboard.

## Steps to reproduce

Hard to reproduce on demand — it needs an application whose accessibility tree blocks. By inspection:

1. `neru launch`, enter hints mode.
2. Trigger a display-configuration change (plug or unplug a monitor, change resolution).
3. The screen-change refresh regenerates hints under the handler lock, calling into the accessibility port with a context carrying no deadline.

The reachable-by-inspection claim is what matters here; the code path is unconditional, not a race.

## Additional context

**ADR 0004 explicitly anticipated this.** It records that the screen-change hints walk "should be bounded when those refreshers move under one hold too" — and #1232 is that move.

**#1232 deliberately did not do it, and the reasoning is worth preserving.** Two reviewers disagreed. The standards review called leaving it unbounded a hard violation of the ADR. `deadlock-reviewer` called it unchanged and not a regression introduced by that PR, and recommended merging without it. The implementer sided with the latter: that walk *already* ran under `h.mu` before the refactor, so the lock hold is exactly as long as it was, and adding `HintTimeout` would be a genuine behaviour change — a hung accessibility tree would exit hints mode rather than block — inside a ticket that promised no behaviour change.

That reasoning is sound for the ticket boundary and is exactly why this is a separate issue rather than a defect in #1241. The behaviour change should be made deliberately, on its own, where it can be described in a changelog entry.

**Scope.** Roughly three lines: bind the context with `HintTimeout` the way the activation path does. Worth checking at the same time whether any other accessibility call reachable under `h.mu` is unbounded, so this is fixed as a class rather than a third instance.

Relevant: `docs/adr/0004-mode-lifecycle-dispatch-under-one-lock-hold.md`, `internal/app/modes/AGENTS.md` (the "don't hold `h.mu` across blocking calls" rule), #1238 for the shape of the fix on the monitor-move path.

y3owk1n:
## Description

This PR fixes a path where an unresponsive application could stop the keyboard.
Regenerating hints after a display configuration change walks the accessibility
tree while the mode handler's lock is held, and that walk had no deadline. The
lock serializes every keystroke, so an application whose accessibility tree
stopped answering held the keyboard for as long as the platform took to give up.

The walk now gets the same five-second budget the activation and monitor-move
paths already give it. A tree that never answers costs hints mode, which exits,
rather than the keyboard. On macOS the walk reads the deadline as it descends,
so the budget cuts a hang short rather than only refusing to start one.

Both bounded walks are now pinned by tests, monitor move included — it was given
this budget when it moved under one lock hold but never had it asserted.

## Related Issues

Closes #1243

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no OS-specific files touched
- [x] No darwin imports from untagged (shared) code — [The One Rule](https://github.com/y3owk1n/neru/blob/main/docs/ARCHITECTURE.md#the-one-rule) — N/A, the change is in shared app code

## General Checklist

- [x] Documentation updated (if applicable) — the modes area guide states the rule, and ADR 0004's deferred bound is marked applied

## Behaviour change

Deliberate, and the reason #1232 left this out rather than doing it silently: a
hung accessibility tree during a display change now ends hints mode after five
seconds instead of blocking. Hints exiting is already what every other failure
on this path does — no elements found, generation failed — so the mode does not
gain a new way to end, only a new reason to reach an existing one. Nothing else
changes: a tree that answers behaves exactly as before, and the budget is the
same one activation has always used.

## Additional Context

**The audit the issue asked for found no third instance to fix, and one thing
worth recording.** Every other accessibility call reachable under the lock is
either already bounded (activation, monitor move, the exclusion check in
validation) or is not a tree query at all — releasing held buttons and
performing an action at a point post a synthesized event, so a deadline buys
them nothing.

**A Go deadline is inert on the focused-application query.** The audit did turn
up an unbounded lookup on the key-dispatch path, and wrapping it was the obvious
move — but it does not work, and the honest version of this PR is that it was
tried and dropped. The accessibility client takes a context, checks it once on
entry, and then makes a synchronous native call; a timeout created microseconds
before that check cannot expire in between, so the wrapper changes no outcome
while costing an allocation on a path that runs per keystroke. Bounding that
call means a bridge-level messaging timeout, which is a different change in a
different layer. The area guide now says which calls need a bound and which
cannot use one, so the next person to notice it finds the answer rather than
re-deriving it.

**Not implemented: `[hints] timeout`.** The issue's config snippet shows a
`timeout` key under `[hints]`, but no such option exists — the budget is a
constant. Making it configurable is a separate decision, and the repo's own
guidance is that every option is a cost, so this PR leaves it alone rather than
inventing one.


y3owk1n:
Valid — fixed in 8d6a7bea.

The assertions captured the context at `FocusedAppBundleID`, which happens before the walk and is the one call on this path the platform makes without reading its deadline. They now capture it at `ClickableElements`, the traversal the budget exists to cut short.

Checked against the exact regression described: handing `ClickableElements` an unbounded context while the earlier lookup stays bounded now fails both tests, where before it passed. Removing the `HintTimeout` wrapper entirely still fails them too.

Issue #1246, opened 2026-08-06
Settle a keymap on change instead of resolving config on every keystroke


y3owk1n:
Spec produced by a grilling session on candidate 4 of `docs/agents/architecture-review-2026-08-05.md`.
Decision record: `docs/adr/0005-the-focused-app-is-published-not-polled.md` — **read it first; it
records five alternatives that were considered and rejected, and the reasons are not obvious from
the code.** Vocabulary: `CONTEXT.md`.

**This is the second of two issues, and it depends on the first.** The measurement issue pins
today's behaviour — one focused-app query per keystroke — and this issue flips that assertion to
zero in the same diff that earns it.

## Problem Statement

While a navigation mode is active and that mode declares per-app hotkey overrides, every keystroke
asks the operating system which application is focused before it can decide what the key is bound
to. The question is asked with the mode handler's lock held.

On macOS the question is an accessibility message to another process, and nothing in Neru sets a
messaging timeout, so it inherits the system default. An application that is busy, hung, or slow to
answer can therefore stall a keystroke — and because the lock is held, it stalls everything else
the handler does, including leaving the mode. The user's escape hatch is behind the same lock as
the thing they are trying to escape.

The cost is also structural. The same merge of base bindings and per-app overrides is computed
three times from the same inputs on the same triggers — once on the keystroke path, twice more when
the modifier-passthrough lists are synchronised — and every one of those recomputes normalises
every binding key in the table. There is no single place that answers "what is bound right now",
which is why the question gets re-answered from scratch each time it comes up.

## Solution

Two changes, and the second is what makes the first honest.

The **focused app is published rather than polled**: the application watcher, which already tells
Neru when the focused application changes, publishes that identifier into a cell the mode handler
owns. Nothing asks the operating system on the keystroke path.

The **keymap is settled rather than rebuilt**: the bindings in force — the current mode's, with the
focused app's overrides applied — are resolved once when the mode, the focused app, or the
configuration changes, and a keystroke consults the result. Modifier-passthrough synchronisation
reads the same settled keymap instead of merging its own copy.

To a user, nothing changes about which key does what, including when they switch applications
mid-mode. What changes is that a wedged application can no longer stall a keystroke, and that
"what is bound right now" has one answer in one place.

## User Stories

1. As a Neru user navigating with hints while a slow application is focused, I want my keystrokes handled without waiting on that application, so that Neru stays responsive when the app I am pointing at is not.
[...]

Issue #1245, opened 2026-08-06
Pin what a keystroke costs inside a mode, before changing it


y3owk1n:
Spec produced by a grilling session on candidate 4 of `docs/agents/architecture-review-2026-08-05.md`.
Decision record: `docs/adr/0005-the-focused-app-is-published-not-polled.md`. Vocabulary: `CONTEXT.md`.

**This is the first of two issues.** It changes no production code. The second one moves the
work off the keystroke path and flips the assertion this issue pins.

## Problem Statement

Neru's guide files say latency is the product — that a correct feature which makes key handling
feel slower is a regression. Nothing in the repository measures key handling. There are two
benchmarks in total, both for grid narrowing, and both run against a fake accessibility port that
answers instantly, so neither can see the cost this issue is about.

That has a concrete consequence. While a mode is active and that mode declares per-app hotkey
overrides, every keystroke asks the operating system which application is focused, with the mode
handler's lock held. On macOS that question is a message to another process. A busy or wedged
application can therefore stall a keystroke, and because the lock is held, it stalls leaving the
mode too. Nobody can currently say how often that happens, how much it costs, or whether a change
made it better or worse — including the change that is about to be proposed for it.

A contributor reviewing that change would have to take a prose claim on faith. A maintainer
revisiting it in six months would have nothing to compare against. And a future change that
quietly reintroduces the same question on the same path would pass every test in the repository.

## Solution

Land the measurement before the fix, and state today's behaviour in code.

After this issue, a keystroke inside hints mode has a benchmark next to the two that already exist
for grid, and a test says out loud how many times a single keystroke asks the operating system
which application is focused — one, when the active mode declares per-app overrides; zero when it
does not. A separate journey states the behaviour that must survive the fix: when the focused app
changes while a mode is active, that app's overrides are the ones in force on the next keystroke.

Nothing a user can observe changes. What changes is that the next issue's diff shows the count
going from one to zero, rather than asserting it in prose.

## User Stories

1. As a contributor reviewing a performance change, I want the count of operating-system calls per keystroke asserted in a test, so that I can see the change take effect instead of trusting the description.
2. As a contributor, I want a benchmark for a keystroke inside hints mode, so that I can tell whether my change to key dispatch cost anything.
3. As a maintainer, I want today's cost recorded before it is changed, so that the improvement has a baseline to be measured against.
4. As a maintainer, I want the benchmark to report allocations as well as time, so that a change which trades allocations for speed is visible as such.
[...]

PR #1249, opened 2026-08-07
docs: name the keymap and focused app, and record how focus is learned

y3owk1n:
## Description

This PR adds two words to the glossary and records one decision. The glossary had
no term for the bindings that are in force at a given moment, nor for the
application the operating system routes keystrokes to, so every discussion of
per-app hotkeys reinvented both — usually as "dispatch table" and "frontmost
app". Those are now **Keymap** and **Focused app**, each with the terms to avoid.

The decision record covers how Neru learns which application is focused.
Resolving a per-mode hotkey currently asks the operating system on the keystroke
that needs the answer, with the mode handler's lock held, and on macOS that
question is a message to another process — so an application that is slow to
answer can stall a keystroke, and the mode exit behind it. The record documents
publishing the focused app from the application watcher and settling the keymap
when the mode, the focused app or the configuration changes, and it lists the
five alternatives that were rejected.

Documentation only — no behaviour changes, and the code the record describes is
not written yet. It is marked **proposed**, and moves to accepted in the change
that implements it.

## Related Issues

Unblocks #1245 and #1246, which both cite this record and this vocabulary.
Closes nothing.

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix
- [x] `docs` — Documentation only

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no code in this PR
- [x] No darwin imports from untagged (shared) code — [The One Rule](https://github.com/y3owk1n/neru/blob/main/docs/ARCHITECTURE.md#the-one-rule) — N/A, no code in this PR
- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no code in this PR

## General Checklist

- [x] `just ci` passes — N/A, docs-only: ran `just fmt-check` and `just lint`, both clean. No Go or Objective-C files changed, so the test, vet, vuln and build stages have nothing in this diff to exercise.
- [x] Tests added/updated for new or changed functionality — N/A, documentation only

## Additional Context

The record's reasoning depends on two facts that are easy to miss when reading
the code, and both are cited in it:

- The accessibility client discards the context it is handed before making the
  native call, so a Go deadline around the focused-application query cannot fire
  — there is nothing between the deadline and the thing that blocks. This is the
  same point the accessibility-deadline rule added to the mode handler's guide
  file in #1248 makes, arrived at independently; the record and that rule agree.
- The application watcher invokes its callbacks inline, under its own read lock,
  and on macOS on the main queue. The mode handler's guide file records that
  nothing running on the main queue may take the handler lock, which is why the
  decision has the watcher publish into a lock-free cell rather than settle the
  keymap itself.

Worth flagging for review: the record draws a line that reads, at first glance,
as a contradiction of ADR 0004 — that record deliberately holds the handler lock
across a whole accessibility tree walk. The line is not "no accessibility work
under the lock" but "a one-shot the user is waiting on may block under the lock
if it carries a deadline; a keystroke may not block at all". If that framing is
wrong, this is the cheapest possible moment to say so, before anything is built
on it.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -34,6 +34,20 @@ _Avoid_: chain, pipeline, script
 A named, parameterised sequence that steps can invoke by name.
 _Avoid_: alias, function, snippet
 
+**Focused app**:
+The application the operating system currently routes keystrokes to. Neru
+names it by the platform's application identifier, and uses it to decide which
+per-app overrides apply.
+_Avoid_: frontmost app, active app, current app, bundle ID as a bare noun
+
+**Keymap**:
+The [[Binding]]s in force right now — the ones the current mode answers to, or
+the global ones when no mode is active, with the [[Focused app]]'s overrides
+already applied. A keymap is settled when the mode, the focused app or the
+configuration changes, not when a key arrives; a keystroke consults one, it
+never builds one.
+_Avoid_: dispatch table, hotkey map, resolved config, keymap table
+
 ### Entering a mode
 
 **Mode command**:
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -45,7 +45,8 @@ The [[Binding]]s in force right now — the ones the current mode answers to, or
 the global ones when no mode is active, with the [[Focused app]]'s overrides
 already applied. A keymap is settled when the mode, the focused app or the
 configuration changes, not when a key arrives; a keystroke consults one, it
-never builds one.
+never builds one. That is the rule the word carries, and the decision behind
+it is ADR 0005; key dispatch still resolves per keystroke until that lands.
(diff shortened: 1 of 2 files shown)


y3owk1n:
Valid catch on `CONTEXT.md` — fixed in 39cd0b44. The entry now says the settling rule is what the term carries and points at ADR 0005, and states plainly that key dispatch still resolves per keystroke until that decision is implemented. The definition itself stays, because "resolved on change rather than per keystroke" is the distinction that makes the word worth having; what was wrong was the present tense reading as a claim about today's code.

PR #1251, opened 2026-08-07
fix(modes): keep a wedged app from stalling a keystroke inside a mode

### LINKED ISSUE #1246: Settle a keymap on change instead of resolving config on every keystroke

y3owk1n:
## Description

This PR fixes a keystroke inside a navigation mode being able to stall on an application that is not answering. Deciding what a key was bound to used to ask the operating system which application is focused, on the keystroke that needed it, while holding the lock that serializes key handling — and on macOS that question is a message to another process with no timeout, so a busy or hung application could freeze the keyboard, mode exit included.

The application watcher now tells Neru which application is focused, and the bindings in force — the current mode's, with that application's per-app overrides applied — are worked out when the mode opens, when the focused application changes, or when the configuration is replaced. A keystroke consults the result and asks the operating system nothing at all.

Nothing changes about which key does what, including switching applications mid-mode: the next key means what the new application's overrides say. Deliberate limitation: where nothing tells Neru about focus changes (Windows, and a Linux session whose compositor exposes no focused-app source), the overrides are learned when the mode opens rather than the instant you switch applications — written down in the capability matrix.

## Related Issues

Closes #1246

## Target Platform

- [x] macOS
- [x] Linux
- [x] Windows

## Type of Change

- [x] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no build-tagged file changed
- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no port method added

## General Checklist

## Additional Context

Decision record: `docs/adr/0005-the-focused-app-is-published-not-polled.md`, moved to **accepted** here. It lists the alternatives that were turned down, including bounding the query with a context deadline — which cannot work, because the accessibility client discards the context before the native call.

**Behaviour worth a second look.** Per-mode modifier passthrough now blacklists the keys of the same settled bindings the keystroke path uses. Previously it resolved per-app overrides for hints only, so grid, recursive grid and scroll blacklisted their base keys and quietly passed a per-app override key through to the focused application. Those modes now consume it, which is what their configuration always said.

**Deviation from the issue.** The issue asks for the cell to be primed once at startup. It is not, and priming would have been wrong: nothing publishes focus changes on Windows, so a primed cell there would pin the keymap to whichever application happened to be focused when the daemon started, and per-app overrides would stop tracking the user (user story 5). Instead the keymap settles when a mode opens — the one-shot the user is already waiting on — which is where the platform is asked if nothing has published yet. That keeps the keystroke path unable to ask on every platform, which is the property the issue is actually after, and it removes a startup call rather than adding one.

**Measurements**, reported as a bonus rather than as the argument — the claim is reliability, not latency. `BenchmarkHintsKeystroke` (hints mode with per-app overrides declared, Apple M3): **13384 ns/op, 3689 B/op, 125 allocs/op → 852 ns/op, 442 B/op, 5 allocs/op**. The benchmark's focused-app query count per keystroke goes from one to zero, and so does the journey that pins it; the journey that switches applications mid-mode and expects the new application's binding to win passes unchanged.

Also here: `go test -race ./internal/app/modes/` covers a publish racing key handling, the `deadlock-reviewer` agent was run on the diff, and the per-mode hotkey accessor that took no focused app (no non-test callers) is deleted.


y3owk1n:
Real finding, and the diagnosis is right — a focus change re-settles the keymap but nothing re-syncs the event tap's blacklist or intercepted-modifier list, so after switching apps mid-mode the keys the tap consumes and the keys the mode is bound to describe different sets.

Two things that change the verdict from blocking to follow-up:

**It is pre-existing.** `syncModifierPassthrough` has always run on mode entry, configuration replacement and the hints refresh, resolving the focused app live at that moment. A mid-mode focus change never re-synced it before this PR either. What this PR changes is the reach — passthrough previously resolved per-app overrides for hints only, and now derives its lists from the settled keymap for every mode.

**The fix is not a call site.** The focused app is published from the application watcher, which invokes callbacks inline and, on macOS, on the main queue. `internal/app/modes/AGENTS.md` records that nothing on the main queue may take the handler lock, so re-syncing from that callback would take the lock and then make an adapter call — the inversion ADR 0005 was written to avoid. It needs the same lazy-settle treatment the keymap got, and the open question is which locked entry point can tell the event tap *before* the next key arrives rather than as a consequence of one.

Tracked as #1252 with a repro, the config that triggers it, and both candidate approaches. Out of scope here deliberately: #1246's spec fixed passthrough synchronisation to the same triggers.

PR #1254, opened 2026-08-07
test(darwin): stop the callback-slot test failing on a busy machine

### LINKED ISSUE #1253: TestCgoSlotConcurrentSetAndDispatch fails on a loaded runner when the scheduler declines the race
## Neru version

`main` at 382bbe0d. The test has been flaky for as long as it has asserted
this; it is not tied to a release.

## Operating system

macOS

## OS version

Seen on the `macos-latest` GitHub Actions runner (also reproducible in
principle on any loaded machine).

## Navigation mode

n.a. — this is a test-suite failure, not runtime behaviour.

## What happened

`TestCgoSlotConcurrentSetAndDispatch` in
`internal/adapter/platform/darwin` failed the `coverage (macos-latest)` job on
a pull request that changes no file in that package:

```
--- FAIL: TestCgoSlotConcurrentSetAndDispatch (0.00s)
    cgo_slot_test.go:144: expected at least one stale snapshot invalidated by concurrent clear
FAIL	github.com/y3owk1n/neru/internal/adapter/platform/darwin	0.292s
```

Re-running the job passed with no code change.

The test asserts `staleSeen.Load() != 0` — that at least one reader observed a
snapshot and found it invalidated before it could check validity. It tries to
make that happen with `runtime.Gosched()` between taking the snapshot and
checking it, while a writer toggles the slot 200 times with `Gosched()` between
each. Nothing guarantees the interleaving. On a loaded runner every reader can
legitimately complete its check before the writer's next toggle, and the test
fails while the code under test is perfectly correct.

## What did you expect

A test of concurrent invalidation should either drive the interleaving
deterministically, or assert a property that holds under every schedule. It
should not fail because the scheduler declined to produce a race the test was
hoping for.

## Steps to reproduce

1. `go test -run TestCgoSlotConcurrentSetAndDispatch -count=200 ./internal/adapter/platform/darwin/`
   on a busy machine, or under `-race` with other work in flight.
2. Observe intermittent failures on the `staleSeen` assertion.

Failure is scheduling-dependent, so a clean run does not disprove it — the
observed CI failure is at
https://github.com/y3owk1n/neru/actions/runs/31143464499/job/92758054644.

## Additional context

Same class as #1247 ("stop the monitor-move lock test failing on a busy
machine"), which is prior art for the fix shape.

Two directions, both better than loosening the assertion to nothing:

- **Drive the interleaving.** Have the reader signal that it holds a snapshot,
  have the writer clear only then, and have the reader check validity after
  that. The race becomes deterministic and the assertion becomes exact rather
  than hopeful.
- **Split the test.** Keep an unsynchronised soak that asserts only what holds
  under every schedule (no torn reads, no dispatch of a cleared value), and add
  a separate deterministic test for the invalidation path.

The value being protected here is real — a stale snapshot must not dispatch —
so the assertion should stay, just stop depending on luck.

The cost of leaving it is paid by unrelated pull requests: this one failed a
green change and needed a manual re-run to clear, which trains people to
re-run red checks without reading them.

y3owk1n:
## Description

This PR fixes a macOS test that could fail on a loaded machine while the code
it covers was perfectly correct. The test checked that a callback holding a
stale snapshot is refused, but it tried to produce the necessary interleaving
by yielding and hoping — so when the scheduler declined the race, the test
failed a green change. That is what happened on CI to an unrelated pull
request, which then passed on a re-run with no code change.

The interleaving is now driven rather than hoped for: the reader hands its
snapshot over, the writer clears only once it is held, and validity is checked
only after that. The race happens on every schedule and every machine, so the
assertion is exact instead of hopeful. The unsynchronised soak stays for its
race-detector value, but now asserts only what holds under every schedule and
reports its counts instead of asserting them.

It also closes a hole in the guarantee the slot exists to make. The dispatch
that runs on another goroutine re-checks validity before firing, because a
clear can land between scheduling the work and running it — but that re-check
had no test, on this branch or on `main`. Deleting it left the whole package
green. The window cannot be forced from outside, so the dispatch body is now a
named method the test can run directly at the moment a clear lands.

Nothing about Neru's runtime behaviour changes.

## Related Issues

Closes #1253

## Target Platform

- [x] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no port or capability changes; the change is confined to darwin-tagged files

## General Checklist

- [x] Documentation updated (if applicable) — N/A, no user-facing surface changes

## Additional Context

**The old failure reproduces, and the new tests survive it.** Loading the
machine with 24 busy processes and running the old test 100 times reproduces
the exact CI failure. The same load against the new tests passes 100 runs
plain and 30 runs under `-race`.

**The positive half of the guarantee is now covered too.** The old test
asserted that at least one dispatch happened, and dropping that assertion
alongside the flaky one would have left every dispatch assertion in the
package negative — a validity check that rejected *everything* would have
passed the whole suite. A separate deterministic test now pins that a snapshot
nothing invalidated does reach its callback, on both the synchronous and
asynchronous paths. Both directions were confirmed by hand-breaking the
validity check each way and watching exactly the right test go red.

**Mutation evidence.** Three ways to break the guard, all three now caught:
making the validity check always accept fails the stale test (and the existing
invalidation test), always reject fails the dispatch test, and deleting the
goroutine-side re-check fails the stale test. That last one is the case this
PR adds coverage for — on `main` it left the entire package green.

**Why `test` and not `fix`.** The prior art named in the issue (#1247) shipped
as a `fix`, but nothing a user can observe changes here, and `test` — like the
`refactor` commit alongside it — keeps this out of the release changelog.
Happy to retype it if you would rather it appear there.

### DIFF
--- a/internal/adapter/platform/darwin/cgo_slot.go
+++ b/internal/adapter/platform/darwin/cgo_slot.go
@@ -93,6 +93,19 @@ func (s *cgoSlot[T]) withValid(callback func(T)) {
 	callback(target)
 }
 
+// dispatchIfValid invokes callback with target when gen is still the live
+// generation. It is the body the async dispatch goroutine runs, named rather
+// than inline so a test can drive it at the moment a concurrent clear lands —
+// the window between scheduling the dispatch and running it is the one this
+// re-check exists to close, and a test cannot force it from outside.
+func (s *cgoSlot[T]) dispatchIfValid(target T, gen uint64, callback func(T)) {
+	if !s.stillValid(gen) {
+		return
+	}
+
+	callback(target)
+}
+
 // withValidAsync invokes callback in a new goroutine when the snapshot is still valid.
 func (s *cgoSlot[T]) withValidAsync(callback func(T)) {
 	target, gen, ok := s.snapshot()
@@ -100,11 +113,5 @@ func (s *cgoSlot[T]) withValidAsync(callback func(T)) {
 		return
 	}
 
-	go func() {
-		if !s.stillValid(gen) {
-			return
-		}
-
-		callback(target)
-	}()
+	go s.dispatchIfValid(target, gen, callback)
 }
--- a/internal/adapter/platform/darwin/cgo_slot_test.go
+++ b/internal/adapter/platform/darwin/cgo_slot_test.go
@@ -77,6 +77,126 @@ func TestCgoSlotWithValidAsyncRejectsStaleGeneration(t *testing.T) {
 	}
 }
 
+// TestCgoSlotDispatchesAValidSnapshot pins the half of the generation guard
+// that says yes: a snapshot nothing invalidated must actually reach its
+// callback, on both dispatch paths.
+//
+// Every other dispatch assertion in this file is negative — "want 0 calls" —
+// so without this one a stillValid that rejected everything would pass the
+// whole package. The async wait is bounded rather than slept on: the timeout
+// is a failure bound, not a race window, so a loaded runner is slow here, not
+// red.
+func TestCgoSlotDispatchesAValidSnapshot(t *testing.T) {
+	var (
+		slot      cgoSlot[int]
+		syncCalls atomic.Int32
+	)
+
+	slot.Set(1)
+
+	slot.withValid(func(value int) {
+		if value != 1 {
+			t.Errorf("withValid dispatched value %d, want 1", value)
+		}
+
+		syncCalls.Add(1)
+	})
+
+	if got := syncCalls.Load(); got != 1 {
+		t.Fatalf("withValid dispatched %d times for a valid snapshot, want 1", got)
+	}
+
+	dispatched := make(chan int, 1)
+
+	slot.withValidAsync(func(value int) {
+		dispatched <- value
+	})
+
+	select {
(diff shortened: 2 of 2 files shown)


y3owk1n:
Good catch — confirmed and fixed in c14427ec.

Verified the gap first: deleting the goroutine-side validity re-check left the
entire `platform/darwin` package green, both here and on `main`, so this was a
real hole rather than one this PR opened.

It could not be closed from the test side alone. The window the re-check guards
is between scheduling the dispatch and running it, which a test cannot force
through the async entry point. So the goroutine body is now a named method, and
the deterministic stale test runs that shared body at the moment a concurrent
clear lands instead of a hand-rolled copy of the same two lines. Production
behaviour is unchanged.

Deleting the re-check now fails `TestCgoSlotStaleSnapshotDoesNotDispatchAcrossGoroutines`.

The "reorders" half of your point still stands: moving the check to before the
`go` statement would pass. I left that uncovered rather than reaching for a
test-only hook in production code.

PR #1320, opened 2026-08-08
test(overlay): hold the macOS label-autohide rule to the shared one

### LINKED ISSUE #1298: refactor(overlay): pin the Objective-C label-autohide copy that ADR 0007 owes a test
## What to build

#1288 collapsed the recursive-grid label-autohide threshold into one shared implementation that Linux and Windows both call. macOS asks the same question in Objective-C (`gridLabelAutohideMultiplier`, around `overlay_darwin.m:1575`) and agrees with the shared answer today — but nothing holds it there.

ADR 0007 is explicit about what is owed in this situation: Go cannot be the one implementation for an Objective-C copy, so the copies get a pinning test instead of a deletion. #1288 named the Objective-C copy in the shared function's doc comment rather than claiming a parity the code does not enforce, and left the test unwritten as out of scope. This ticket writes it.

Worth doing alongside #1289, which builds the same kind of guardrail for the hint-placement vocabulary — the two are the only instances of the language-boundary exception in the codebase, and whatever shape the first one takes, the second should match.

## Acceptance criteria

- [ ] A test fails if the Objective-C label-autohide rule drifts from the shared Go implementation — same treatment of a non-positive multiplier, same threshold, same comparison against both cell dimensions
- [ ] The test is verified to have teeth by deliberately breaking the Objective-C side and watching it fail
- [ ] Its shape is consistent with the guardrail built for the hint-placement vocabulary in #1289
- [ ] `just ci` passes

## Blocked by

None — can start immediately.

y3owk1n:
## Description

This PR adds a guardrail so the macOS overlay cannot quietly start hiding
different recursive-grid key labels from Linux and Windows. Whether a cell is
big enough for its label is one rule — a multiplier times the label font size,
which both cell dimensions have to reach, with a non-positive multiplier
meaning "always show" — and it is written twice: once in shared Go that the
Linux and Windows overlays call, and once in Objective-C for macOS. The two
agreed, but nothing held them there, so a change to either would have shown one
person a labelled cell and another a blank one from the same configuration.

The Objective-C rule is now read out of the source and run against the same
questions as the shared one, and any answer they give differently fails the
build. Nothing changes on screen and no option changes meaning.

One deliberate limitation: the guardrail reads the Objective-C rule by its
shape, so rewriting that code in a way that keeps the behaviour fails the test
until the guardrail is taught the new shape. That is the cost of pinning a rule
that lives in another language, and the failure names exactly what it expected
to find.

## Related Issues

Closes #1298

## Target Platform

- [x] macOS
- [x] Linux
- [x] Windows

## Type of Change

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no new Go file is build-tagged; the guardrail reads the macOS source as text so it runs on every host.
- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no port or capability changed.

## General Checklist

- [x] Documentation updated (if applicable) — ADR 0007 records this pin as landed, and its stale line about earlier work is corrected; no user-facing doc changed, because no behaviour or option did.

## Additional Context

This is the second of the two language-boundary pins ADR 0007 owes, after the
hint-placement vocabulary. That one pinned constants, which two small
extractors could read; this copy is a comparison rule with a runtime property,
so no constant extractor could be aimed at it. It brings its own reader
instead, and shares the native-source entry point with the first pin.

Nothing in the test is a hand-written expectation. The Objective-C rule is
parsed into something the test can evaluate, and the shared Go implementation
is the only expectation — which makes the pin bidirectional: change either side
alone and the two stop answering alike.

The teeth were verified two ways. By hand, seven breaks to the real Objective-C
source — the comparison loosened, one of the two dimensions dropped, the
multiplier check inverted, its bound raised so smaller multipliers stopped
hiding anything, the threshold taken from another font, the method renamed, and
the guard folded into the comparison — each failed with a message naming the
case and the two sides. And durably, in the test file itself: one test mutates
the rule it read and fails if no case can tell the mutant apart from the shared
implementation, so the case list cannot rot into one that agrees with
everything. The raised-bound break is the one that first slipped through, and
the case that catches it was added because of it.

One part of the rule is pinned by shape rather than by behaviour, and the test
says so: with a cell size and a font size both non-negative, deleting the
multiplier guard outright answers identically to keeping it on every input
either implementation can be handed. Reading the guard's presence is what
catches that.

`just lint-cross` was run alongside `just ci`, since the guardrail package now
compiles against a renderer that has a macOS-specific file.

Two things noticed and deliberately left alone. The sub-key preview autohide
threshold is the same shape of language-boundary copy — Linux Go and macOS
Objective-C measure a sub-cell exactly the same way — and it has no pin; ADR
0007 does not name it, so this PR does not claim the boundary is now fully
covered. (Windows measuring the whole cell instead is a documented, deliberate
difference, not a copy that drifted.) And ADR 0007's context and
considered-options sections still describe the pre-#1292 state in the present
tense; that is the situation the decision was taken in, so past-tensing it
would rewrite the record rather than correct it — only the consequences
section, which tracks landed-versus-owed work, is updated here.

### DIFF
--- a/docs/adr/0007-a-shared-derivation-has-one-implementation.md
+++ b/docs/adr/0007-a-shared-derivation-has-one-implementation.md
@@ -80,8 +80,13 @@ are the precedent, not the exception.
   and the `HintPlacement` enum to the same numbering. It pins *which placement
   each value means*, not where the badge is then drawn — `hintRectForPlacement:`
   remains a single Objective-C implementation with no Go counterpart to
-  disagree with. One copy of this kind is still owed a pin: the recursive-grid
-  label-autohide rule (#1298).
+  disagree with. The second such pin landed with the recursive-grid
+  label-autohide rule (#1298), where the copy is a rule rather than a constant:
+  `internal/architecture/label_autohide_rule_test.go` reads the Objective-C
+  guard in `drawGridLabel:` into something it can run, and holds its answer to
+  `recursivegrid.Style.ShowLabelIn` over the cases that separate them — the
+  multiplier that disables autohide, the threshold itself, and each cell
+  dimension one pixel under it.
 - **"Lowest layer with more than one caller" is a judgement someone has to
   make, and it moves.** A derivation with one caller is not shared and should
   stay private; the second caller is what triggers the move, and the move is
@@ -90,9 +95,11 @@ are the precedent, not the exception.
   which is why `CONTEXT.md` lists *helper*, *util*, *common code* and *shared
   logic* as the words to avoid. A shared derivation is named for what it
   derives.
-- **This ADR ships one instance and leaves two known ones standing.** The font
+- **This ADR shipped one instance and left two known ones standing.** The font
   parser is collapsed here (#1286). The subgrid breakpoints and the 3×3 written
-  five times are filed separately (#1287), because they touch the drawing path on every
-  backend and deserve their own diff. Naming the rule before finishing the work
+  five times were filed separately (#1287), because they touch the drawing path on every
+  backend and deserved their own diff; they were collapsed in #1292, where the
+  mode layer and all three overlays came to ask `internal/domain/grid` for the
+  cells rather than divide the rectangle themselves. Naming the rule before finishing the work
   is the point: the next platform backend should not have to infer it from two
   packages that never explained themselves.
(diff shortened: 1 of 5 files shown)


y3owk1n:
~~Heads up on the red checks: they are not from this branch.~~ **Resolved — this is stale, kept for the record.**

The red checks came from `main`, not this branch: #1317 landed a call to `isSkippedWalkDir(entry.Name())` in `internal/architecture/foundation_slice_test.go` while #1316 had changed that helper to take `(repoRoot, path)`, so every lint, vet and test job failed on main itself and this PR inherited it through the merge ref.

#1322 fixed it. This branch is now rebased onto `4adfc3e2`, and `just ci`, `just test-foundation` and `just lint-cross` all pass locally. Nothing was needed on this side beyond the rebase.

PR #1330, opened 2026-08-08
fix(overlay): keep odd-sized badges at their full size

### LINKED ISSUE #1326: Linux and Windows draw hint badges a pixel smaller than macOS when the size is odd
## What to build

The same hint badge is one pixel narrower and one pixel shorter on Linux and Windows than on macOS, whenever its width or height is odd.

macOS computes the badge's origin in floating point and draws the full size — `targetX - boxWidth / 2.0`, with the rect then spanning `boxWidth` — so a 41-pixel badge is 41 pixels wide. Linux and Windows compute both edges with integer halves, `center ± size/2`, so a 41-pixel badge spans 40. Even sizes agree; odd sizes never do. The same split applies to the hint boundary highlight, which is derived the same way on each side.

Nobody has reported it, and at one pixel nobody is likely to. It is worth fixing anyway because it is a rule, not an accident: every future badge-shaped thing inherits whichever form its platform happens to use, and "the overlays agree except when a number is odd" is the kind of caveat that quietly becomes load-bearing.

**This got cheap to fix.** Before #1325 the arithmetic was written at five sites across two backends. It is now a single Go function, `badge.CenteredOn` — so the decision is one place, and there is already a test asserting exactly how the two forms diverge (same near edge, far edge one pixel short, only on odd sizes).

The decision itself is which rounding wins. Matching macOS means changing `CenteredOn` to span the full size from a halved origin, which moves badges by a pixel on Linux and Windows. Matching the integer form means changing the Objective-C, which moves them on macOS. macOS is the reference platform and the float form loses nothing, so it is the more likely answer — but it is a deliberate pixel shift on two platforms either way, and it belongs in the changelog as one.

Whichever wins, the loser's arithmetic should stop existing rather than being left as a second correct-looking form for someone to reintroduce.

## Acceptance criteria

- [ ] A hint badge of a given size occupies the same rectangle on macOS, Linux and Windows, for odd and even sizes alike
- [ ] The hint boundary highlight agrees the same way
- [ ] One rounding rule exists; the rejected form is gone rather than left beside it
- [ ] The pixel shift is stated in the PR body for the changelog, naming which platforms move
- [ ] A test covers odd sizes specifically — that is the only place the two forms ever differed
- [ ] `just ci` and `just lint-cross` pass

## Blocked by

None — can start immediately.

Guflly:
## Description

This PR keeps odd-sized hint badges and target highlights at their requested dimensions on Linux and Windows. Their right and bottom edges now extend by one pixel when needed, matching macOS while leaving even-sized geometry unchanged.

The same size-preserving centering rule now applies wherever a rectangle is centered on a point.

## Related Issues

Closes #1326

## Target Platform

- [ ] macOS
- [x] Linux
- [x] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

## General Checklist

- [x] Documentation updated (if applicable) — N/A, no documentation surface changed

## Screenshots / Recordings

Not included. The one-pixel difference is covered by exact rectangle assertions.

## Additional Context

Odd-sized hint badges and target highlights gain one pixel on their right and/or bottom edge on Linux and Windows. Even sizes do not move. Linux monitor-selection panels use the same rule and preserve their configured odd width as well.

### DIFF
--- a/internal/adapter/overlay/linux/hint_arrow_test.go
+++ b/internal/adapter/overlay/linux/hint_arrow_test.go
@@ -161,22 +161,16 @@ func TestHintBadgePlacement_ArrowBaseClampedToFlatEdge(t *testing.T) {
 	}
 }
 
-// TestHintBadgePlacement_OddBadgeSizeDropsItsLastPixel pins the rounding the
-// badge is placed with. The badge hangs on its target point as center ± half,
-// so a badge sized to an odd width or height is drawn one pixel short of that
-// size. The alternative rounding (badge.CenteredIn's center - half, + size)
-// would keep that pixel and move the badge's right and bottom edges, so this
-// test is what stops the two being swapped.
-func TestHintBadgePlacement_OddBadgeSizeDropsItsLastPixel(t *testing.T) {
+func TestHintBadgePlacement_OddBadgeSizeKeepsItsSize(t *testing.T) {
 	target := image.Pt(100, 100)
 
 	centered, _, _ := hintBadgePlacement(target, 41, 21, 4, config.HintPlacementCenter)
-	if want := image.Rect(80, 90, 120, 110); centered != want {
+	if want := image.Rect(80, 90, 121, 111); centered != want {
 		t.Errorf("center placement badge = %v, want %v", centered, want)
 	}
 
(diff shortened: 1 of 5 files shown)


y3owk1n (approved):
Looks good, we'll merge when ci goes green. Thanks!

PR #1355, opened 2026-08-08
fix(app): notice the first application switch after the daemon starts

### LINKED ISSUE #1348: test(app): TestSimulation_FocusChangeMidModeRebindsTheKey is flaky, and the obvious cause is not the cause
## What to build

`TestSimulation_FocusChangeMidModeRebindsTheKey` fails intermittently. Observed on PR #1347, CI run `31260179436`, **attempt 1**, job `test (ubuntu-latest)`:

```
simulation_journey_test.go:2182: pressing "j" after switching to Other App entered scroll,
    want recursive_grid: the key still means what it meant in the application the mode was
    opened in
--- FAIL: TestSimulation_FocusChangeMidModeRebindsTheKey (0.01s)
```

Attempt 2 — a rerun of the same commit, no code change — passed all 22 checks, and `main` was green across the preceding runs. So this is intermittent rather than a regression: the PR that surfaced it touched only `.gitignore` and `internal/architecture/`, neither of which can reach `internal/app`.

What the failure means: at `HandleKeyPress` time the settled keymap still carried the *previous* application's per-app override, so the shared key ran `scroll` (the fixture app's binding) instead of `recursive_grid` (the binding of the application the user had just switched to).

**Note what the cause is not.** The obvious reading — that the test is missing a `waitFor` between `sim.focusApp(...)` and `sim.press(...)`, the way the sibling test below it at `internal/app/simulation_journey_test.go:2203` has one — does not survive reading the path:

- `simHarness.focusApp` sets the fixture's focused app and then calls `MockAppWatcherPort.EmitActivate`, which invokes each registered callback **inline** (`internal/ports/mocks/appwatcher.go:125`).
- `App.handleAppActivation` calls `modes.PublishFocusedApp` as a synchronous, lock-free cell write (`internal/app/lifecycle.go:307`).
- `settledKeymap` re-settles on the next read, and `focusedApp` is part of the `keymapInputs` comparison (`internal/app/modes/keymap.go:98-111`).

Publication and the keystroke therefore happen on the same goroutine, in order, with no publish/read gap to wait on. The sibling test's `waitFor` exists because the *passthrough blacklist* refresh is asynchronous — a different concern from the keymap.

That points at the asynchronous work the same activation kicks off as the place to look: `go a.modes.RefreshPassthroughForFocusedAppChange()` and `go a.hotkeys.RefreshFor(bundleID)` (`internal/app/lifecycle.go:316` and `:319`), possibly still in flight from an earlier activation, or something else re-resolving the focused app behind the settle.

**Diagnose before fixing.** A `waitFor` added to quiet the test would also hide a real ordering bug in the activation path — where a keystroke genuinely can arrive before the activation is absorbed, and the user gets the wrong mode.

## Acceptance criteria

- [ ] The cause is identified and stated in the fix, not papered over by a wait that hides it
- [ ] `go test -run TestSimulation_FocusChangeMidMode -count=200 -race ./internal/app/` passes
- [ ] If the cause turns out to be in the activation path rather than the test, the production fix lands with it
- [ ] `just ci` passes

## Blocked by

None — can start immediately.

y3owk1n:
## Description

This PR fixes the daemon missing the first application switch after it starts. It began watching for switches, and accepted its first hotkey, before it had registered the handler those switches are delivered to — so a switch reported in that window reached nobody, and nothing goes back for one that was missed. On Linux the effect lasts longest: the watcher takes a single sample as it starts and reports only changes after that, so the sample that went missing is not re-reported until the user switches application again.

What that cost the user is per-app hotkey overrides binding to the wrong application. While nothing has been reported, the bindings in force are settled against whichever application was focused when the mode was opened, so a key pressed after switching ran the meaning it had in the application they had just left. Startup now registers the handler first, brings the hotkeys up next, and starts the watcher last, which also stops a switch arriving mid-startup from having its per-app hotkey registration overwritten by the daemon's own initial one.

The window is narrow and only reachable in the first moments after launch, so most users would have seen this as an occasional oddity rather than a reproducible one. It is what made `TestSimulation_FocusChangeMidModeRebindsTheKey` fail intermittently on CI.

## Related Issues

Closes #1348

## Target Platform

- [x] macOS
- [x] Linux
- [x] Windows

## Type of Change

- [x] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no OS-specific file is added or changed.
- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no port contract changes.

## General Checklist

## Additional Context

Three things worth a reviewer's attention.

**Why the startup order is the cause, and not the test.** The issue ruled out the obvious reading — a missing wait between the simulated focus change and the keystroke — and it was right to: publication is a lock-free write on the same goroutine as the keystroke, so there is no publish/read gap. It is not either of the two goroutines the issue pointed at either; one only reads, and the other is deferred outright while a mode is open. Injecting a delay between the daemon making its hotkeys live and registering its watcher handler reproduces the reported failure verbatim on the first run, and the fix makes it pass under a 300ms delay in the same place.

**The simulation harness was waiting for nothing.** It gated on the app reporting itself enabled, which is true before the startup sequence is entered at all, so journeys could drive the fixture desktop before the daemon was listening. It now waits for the watcher to be running — the last of the three steps the fix orders — which is a real barrier rather than a wait that would have hidden the bug. The counterfactual holds: with that gate in place but the production order reverted, the new pin test still fails.

**The rest of the old startup block stayed where it was.** Only the event registrations moved ahead of the watcher; the theme observer and Mission Control gating did not, because the theme observer reaches a session bus on Linux and a bus round trip in front of the watcher and the hotkeys would delay the daemon accepting its first chord.

The change also adds a pin test asserting the order directly rather than the symptom, since whether the window is wide enough to fall into is a property of the machine. ADR 0005 gains the consequence that made the ordering load-bearing: publishing has one delivery and no retry, where polling was self-healing.

### DIFF
--- a/docs/adr/0005-the-focused-app-is-published-not-polled.md
+++ b/docs/adr/0005-the-focused-app-is-published-not-polled.md
@@ -74,6 +74,17 @@ exit included, since it takes the same lock.
   about focus instead of asking. One `resolveFocusedApp` covers both: read the
   cell when it is fed, ask when it is not. `docs/CROSS_PLATFORM.md` gains the
   row.
+- **Publishing only works while something is listening, so the order the daemon
+  starts in became load-bearing.** Polling was self-healing: whoever asked got
+  an answer, however late they asked. A publication has one delivery and no
+  retry, so a watcher started before `App.Run` registers its activation callback
+  drops what it reports, and on Linux — where the watcher samples once as it
+  starts and reports only changes after that — the lost sample is not
+  re-reported until the user switches application again. Until then the keymap
+  settles by asking, which is what this ADR set out to avoid, and per-app
+  overrides bind to whatever was focused when the mode opened. #1348 is that
+  bug; `app/lifecycle.go` registers before it starts, and the simulation journey
+  pins the order.
 - **`handlerState.focusedBundleID` is deleted.** Leaving it would leave the
   regression writable; deleting it means the only way to learn the focused app
   inside the handler is the published cell. Callers that legitimately need a
--- a/internal/app/lifecycle.go
+++ b/internal/app/lifecycle.go
@@ -69,15 +69,38 @@ func (a *App) Run() error {
 
 	a.logger.Info("IPC server started")
 
-	a.appWatcher.Start()
-	a.logger.Info("App watcher started")
+	// These three are ordered, not merely sequential, and the order is starting
+	// the watcher last: an activation it delivers reaches both of the things
+	// above it, and neither absorbs one until it has run. Registering after the
+	// watcher started meant a dropped activation, which nothing retries — ADR
+	// 0005 has the consequence, #1348 was the bug, and
+	// TestSimulation_TheAppHearsFocusChangesFromTheMomentItWatchesForThem pins
+	// it. Registering the hotkeys after it would hand the same activation a
+	// binder holding nothing, and the refresh here would then re-register the
+	// global table over the per-app one it had just installed
+	// (keybinding/binder.go).
+	//
+	// Only the registrations belong ahead of this; nothing that can block does.
+	// The observers below reach a session bus on Linux, and putting a bus round
+	// trip in front of the watcher and the hotkeys would delay the daemon
+	// accepting its first chord to hurry up something nothing is waiting on.
+	a.registerAppWatcherCallbacks()
 
 	a.hotkeys.RefreshFor("")
 	a.logger.Info("Hotkeys initialized")
 
-	a.setupSleepObserver()
+	a.appWatcher.Start()
+	a.logger.Info("App watcher started")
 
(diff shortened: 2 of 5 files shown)


y3owk1n:
Fixed — that was a regression I introduced in this PR rather than pre-existing behaviour: the line previously read a fresh snapshot and I narrowed it to the one taken at the top of startup while moving it. Restored, with a note on why it has to be read there.

Worth flagging separately: nothing re-applies this setting on a config reload at all, so toggling `hints.detect_mission_control` still needs a restart to take effect. That predates this PR and I have left it alone here.

PR #1391, opened 2026-08-09
fix: only excuse a failed flush of output that is provably gone

### LINKED ISSUE #1380: test(app): race-leg teardown prints ~128 'failed to close logger' warnings that bury real failures
## What to build

Every `just ci` race-test leg emits roughly 128 lines of `Warning: failed to close logger: [LOGGING_FAILED] ... sync /dev/stderr: bad file descriptor` during test teardown. The warnings are benign — syncing stderr is expected to fail once the test binary's descriptors are being torn down — but at that volume they make grepping gate output for real failures needlessly hard, and they train readers to ignore warnings.

Make logger close/sync during test teardown treat `sync /dev/stderr`-class errors as expected (the standard zap-on-stderr caveat) so the warning only fires on genuinely failed sinks. Observed 2026-08-09 on the `just ci` runs for PRs #1377/#1378.

## Acceptance criteria

- [ ] A full `just ci` run emits zero `failed to close logger` warnings from test teardown
- [ ] A genuinely failing logger sink still warns
- [ ] `just ci` passes

## Blocked by

None — can start immediately.

y3owk1n:
## Description

This PR tightens the shutdown-warning fix from #1388, which merged while a second automated review comment on it was still open.

That change stops warning about a flush of Neru's own console output when the destination is something a flush could never have worked on. Deciding that required inspecting the destination — and when the inspection itself failed, for any reason, the destination was assumed unflushable and the failure was dropped. Redirected output that failed to flush for an unrelated reason could therefore be silently discarded. Now only a descriptor that is provably gone earns that silence, which is what the behaviour was documented as doing; anything else is reported.

In practice this is a very narrow window — inspecting an open descriptor essentially only fails when it is gone — so this is a correctness tightening rather than a fix for something users have hit.

## Related Issues

Follow-up to #1388 (which closed #1380).

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no platform-tagged files

## General Checklist

- [x] Documentation updated (if applicable) — N/A, no user-facing option, command or documented behaviour changed

## Additional Context

A test pins the remaining silent case: a stream whose descriptor is already closed is still treated as expected noise, because there is nothing left to flush to. A full `just ci` run on this branch exits 0 and still emits zero `failed to close logger` lines, so the original noise fix is intact.

### DIFF
--- a/internal/adapter/logger/syncerror.go
+++ b/internal/adapter/logger/syncerror.go
@@ -81,11 +81,12 @@ func isUnflushableStreamError(err error, streams []*os.File) bool {
 // redirected, so the name alone says nothing: a redirect to a regular file is
 // flushable, and a failure there is real and must be reported. A stream whose
 // descriptor is already gone counts as unflushable — there is nothing left to
-// flush it to.
+// flush it to — but a stream that cannot be inspected for any other reason is
+// given no benefit of the doubt.
 func isUnflushable(stream *os.File) bool {
 	info, err := stream.Stat()
 	if err != nil {
-		return true
+		return errors.Is(err, syscall.EBADF) || errors.Is(err, os.ErrClosed)
 	}
 
 	return !info.Mode().IsRegular()
--- a/internal/adapter/logger/syncerror_test.go
+++ b/internal/adapter/logger/syncerror_test.go
@@ -96,6 +96,24 @@ func TestGenuineSyncFailure_ReportsFailureOnRedirectedRegularFile(t *testing.T) 
 	}
 }
 
+// A stream whose descriptor is already gone has nothing left to flush to, even
+// when it was pointing at a real file.
+func TestGenuineSyncFailure_NilForClosedStream(t *testing.T) {
+	stream := regularFileStream(t)
+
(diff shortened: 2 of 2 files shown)


y3owk1n:
Closing this: CI shows the tightening is not portable.

The premise was that a stream which cannot be inspected should be reported rather than excused, and that "already gone" could be recognised from the inspection error. On Windows it cannot — `Stat` on a closed stream reports something that is neither of the two "descriptor is gone" signals, so the branch that is supposed to stay quiet starts warning again (test (windows-latest) here, on the test that pins exactly that case).

That leaves "cannot prove the destination is a real file" as the only portable rule, which is what main already does. The case this PR was aimed at — output redirected to a real file that fails to flush — is still reported, because inspecting an open descriptor succeeds and reports a regular file; the residual window is an inspection that fails for a reason other than the descriptor being gone, which in practice does not happen. Leaving main as it is.

PR #1404, opened 2026-08-09
test(architecture): hold every Linux tap to one spelling of the keypad

### LINKED ISSUE #1402: test(architecture): the Wayland keypad fold table has three Go transcriptions and none is pinned to it
## What to build

`neru_normalize_xkb_name` in `internal/adapter/platform/linux/wayland_keymap.c:186-258` decides what each keypad keysym is called. That decision is now transcribed into Go three times, and nothing holds any copy to the C source:

- the keysym cascade in `internal/adapter/eventtap/linux/x11_cgo.go` (#1385 / PR #1395)
- `evdevKeyNames` in `internal/adapter/eventtap/linux/evdev_keys.go` (#1399 / PR #1400)
- the test tables mirroring both

Both review passes on PR #1400 flagged this independently. It is exactly what ADR 0007 asks for — where the second implementation is in another language, Go cannot be the one implementation, so the copies get a test holding them together. The three copies agree today because each was written by reading the previous one; nothing makes them keep agreeing, and a disagreement means one physical key producing different names on different Linux backends.

The machinery exists: `internal/architecture/named_key_tables_test.go` (#1374 / PR #1396) reads Objective-C tables as text through the shared `readNativeSource` helpers and holds them to `keyvocab`. This is the same shape one language over.

The real difficulty, and why #1395 and #1400 both deferred it: `internal/architecture` is untagged, so it cannot reference `evdevKeyNames` or the X11 cascade directly — the pin has to parse both Go sources as well as the C one. Budget for that rather than assuming the macOS pin's shape transfers unchanged.

Add the pin to ADR 0007's inventory paragraph in the same change, as that inventory requires.

## Acceptance criteria

- [ ] An architecture test fails if any of the three Go transcriptions disagrees with `wayland_keymap.c` about what a keypad keysym is called
- [ ] The test reports a table it cannot parse rather than silently covering less than it claims — the failure mode #1396 had to fix in review
- [ ] A negative meta-test proves the pin still catches drift
- [ ] ADR 0007's pinned-copy inventory names the new pin
- [ ] `just ci` passes

## Blocked by

None — can start immediately.

y3owk1n:
## Description

This PR adds a guardrail that keeps every Linux backend calling the keypad by the same names. What the keypad reports is decided once, in the Wayland keymap's C fold table, and that decision had been retyped into Go three times — the X11 keysym lookup, the evdev fallback table, and the tests mirroring both. They agreed because each was written by reading the previous one; nothing made them keep agreeing, and a name retyped on one side would have had one physical key reach different bindings depending on which Linux tap is running.

The new architecture test reads all four copies and holds them to the C source. Each keypad keysym declares which copies must carry it, so the deliberate gaps fail the day they stop being gaps, and a copy written in a shape the test cannot parse is reported rather than skipped. One limitation is stated in the test itself: the keypad's decimal key is excused from both Go copies because no Go table spells the name the keymap gives it, so that one row is held for presence and not for agreement.

No production code changes — the three copies agree today, and the test documents that they do.

## Related Issues

Closes #1402

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no platform-specific file is added or changed; the guardrail is untagged and reads the tagged ones as text.
- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no port or platform capability changes.

## General Checklist

## Additional Context

The reason the two earlier keypad PRs deferred this: `internal/architecture` is untagged, so it cannot reference either Go copy — one is behind `//go:build linux` and the other behind cgo as well. Both are therefore parsed as text with `go/ast`, the way the hint-placement pin already reads one build-tagged Go function, while the C table is read through the same reader every language-boundary pin here shares.

Three things the pin is deliberate about, all of them lessons from the macOS key-table pin:

- Every table is read in full or reported. A row, entry, case arm or test row written in a shape the parser does not know fails with the shape it expected, rather than being skipped while the pin reports coverage of it. A dedicated test doctors each of the five sources to prove that.
- Deliberate absences are asserted in both directions. A keysym the X11 lookup skips because Xlib answers it with a character fails if a case appears for it, exactly as a missing case fails.
- A drift meta-test mutates the parsed tables seventeen ways — an entry dropped, a name retyped, a test row agreeing with the copy it mirrors and nothing else — and requires each mutant to fail a check.

Scope is the keypad rows. The same C table also renames punctuation keysyms, `Page_Up`/`Page_Down` and `Caps_Lock`, and those are read but not pinned: the evdev table answers by kernel key code, and only for the keypad does the tree say which keysym a code reports. The test and the ADR both say so rather than implying wider coverage.

ADR 0007's inventory of pinned language-boundary copies names this pin, as that inventory requires, and the shared native-source reader's doc comment now says it reads C functions as well as Objective-C methods, since this is the first pin to use it on one.

### DIFF
--- a/docs/adr/0007-a-shared-derivation-has-one-implementation.md
+++ b/docs/adr/0007-a-shared-derivation-has-one-implementation.md
@@ -102,8 +102,8 @@ are the precedent, not the exception.
   predicate. The pin now runs the shared Go rule and reads only the Objective-C
   one, exactly as the label-autohide pin beside it does.
 - **The exception is the half of this rule deletion cannot enforce, so what is
-  pinned is inventoried here.** Five language-boundary copies are pinned as of
-  #1374: the three named above, plus the synthetic key-up and modifier-toggle
+  pinned is inventoried here.** Six language-boundary copies are pinned as of
+  #1402: the three named above, plus the synthetic key-up and modifier-toggle
   wire prefixes, which `internal/adapter/platform/darwin/eventtap_darwin.m` and
   `internal/adapter/platform/linux/overlay_wayland.c` format with `printf`
   because neither can import `internal/domain/keyvocab`, held to that package by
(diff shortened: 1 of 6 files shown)


y3owk1n:
Both Greptile findings were real and are fixed in dc8c1f1e: a keysym folded twice is now reported rather than read as its later row (which the C lookup never reaches), and the KP_0-KP_9 rule is read with its `blen == 4` guard, so dropping that guard fails instead of passing as intact. Each is covered by doctoring the C source and requiring the pin to report.

Issue #1406, opened 2026-08-09
docs(darwin): the callback contract mandates an idiom fifteen packages cannot reach


y3owk1n:
## What to build

`internal/adapter/platform/darwin/AGENTS.md` states: *"C→Go callbacks funnel through `cgoSlot[T]` — never a raw package var."* `cgoSlot` is unexported. Every one of its six users is inside the package, and the fifteen packages that include a bridge header cannot obey the contract they are held to.

The contract is what is wrong, not the visibility. The two cases are genuinely different problems:

- `cgoSlot[T]` tracks **one long-lived handler per subsystem**. The C side passes no context; the `//export`ed function finds the process-global slot; the generation counter invalidates when the handler is replaced.
- `overlayutil.CallbackManager` tracks **N in-flight one-shot operations**. `NeruResizeOverlayToActiveScreenWithCallback` hands its context through two nested `dispatch_async` hops, so each operation needs its own identity, allocated on the C heap because Go's GC cannot see it, plus a timeout because the callback may never arrive at all.

`cgoSlot` structurally cannot serve the second — it has no per-operation identity to hand C. So exporting it would offer packages a tool that does not fit their case and invite exactly the wrong reuse. ADR 0009 (`docs/adr/0009-a-bridge-interface-is-its-headers.md`) records that rejection; this ticket lands the wording it implies.

Restate the bullet as two idioms, each with the case it serves: callbacks registered inside the bridge funnel through `cgoSlot[T]`, which stays unexported; callbacks that cross an async dispatch back into another package go through the callback registry, whose context crosses on the C heap. The existing `MallocCallbackContext` / `FreeCallbackContext` sentence is about the second idiom and should read as part of it rather than as a footnote to the first.

Documentation only — no code changes, and `cgoSlot` is not exported. Deliberately not pinned by a test: what the old wording reached for — no raw package var holding a C callback target — is not statically checkable, since deciding which package vars are callback targets means following `//export` reachability, and every approximation either misses the case it exists for or fires on innocent vars. The contract carries this one.

`platform/darwin/CLAUDE.md` is a symlink to `AGENTS.md`; edit the guide file, never the symlink.

## Acceptance criteria

- [ ] `internal/adapter/platform/darwin/AGENTS.md` names two callback idioms and the case each one serves
- [ ] The restated contract no longer mandates `cgoSlot[T]` to packages outside the bridge
- [ ] `cgoSlot[T]` remains unexported and no code changes
[...]

Issue #1405, opened 2026-08-09
test(architecture): the overlay render tree reaches the darwin bridge only by accident


y3owk1n:
## What to build

`internal/adapter/platform/darwin` publishes its Objective-C as headers, and fifteen packages call into it by relative `#include`. Including a header is not what links the compiled objects into the binary — importing the Go package is. ADR 0009 (`docs/adr/0009-a-bridge-interface-is-its-headers.md`) records the rule that follows: a bridge's interface is its headers, its Go import links them, and every consumer states that import directly.

Six packages already do. One needs no archive at all — `internal/adapter/systray/darwin` compiles its own `.m`, so it needs only the header. The remaining eight are the entire overlay render tree, and they reach the archive only transitively, all of them through `overlay/render/overlayutil/native` — a three-function shim none of them names:

- `internal/adapter/overlay/darwin`
- `internal/adapter/overlay/render/grid`
- `internal/adapter/overlay/render/hints`
- `internal/adapter/overlay/render/modeindicator`
- `internal/adapter/overlay/render/overlayutil`
- `internal/adapter/overlay/render/recursivegrid`
- `internal/adapter/overlay/render/stickyindicator`
- `internal/adapter/overlay/render/virtualpointer`

That is an accident holding up a stack. Refactor `overlayutil` to stop needing `native` — a plausible change — and all eight fail to link at once, on macOS only, with an undefined-symbol error naming none of them. It is the same cost `cgo_includes_test.go` already documents for the other half of this boundary: `go vet` does not see it, the cross-platform vet does not see it, and it surfaces as a red CI job minutes later.

Give the eight a direct import of the bridge, and add a guardrail that keeps it that way: a package that includes a bridge header and carries no native source of its own must import that bridge package. It belongs beside `TestCgoIncludes_NativeBridgePointsAtThePlatformPackages`, which pins where a header may be reached *from*; this pins that the objects behind it are actually linked. The `systray/darwin` case is encoded as part of the rule rather than special-cased by name — it is what shows the rule is about linkage and not about includes.

[...]

PR #1410, opened 2026-08-09
test(architecture): pin the macOS callback context struct to its Go twin

### LINKED ISSUE #1407: test(architecture): the C callback context struct is pinned to its Go twin by a comment
## What to build

`internal/adapter/platform/darwin/callback_context.h` declares a two-field struct and documents it as *"matches overlayutil.CallbackContext in Go"*. That match is load-bearing: the resize completion path allocates the struct on the C heap, native code retains it across two `dispatch_async` hops, and the `//export`ed Go side casts the raw pointer straight to `overlayutil.CallbackContext` and dereferences it. If the two layouts disagree, that read is silently wrong — a callback ID and generation pulled from the wrong offsets, validated against a registry, and dropped or dispatched on garbage.

Nothing holds them together but the comment, and the two halves live in packages that cannot see each other.

This is a missed case of a convention that already runs, not a new one. ADR 0007 (`docs/adr/0007-a-shared-derivation-has-one-implementation.md`) already licenses a native copy of a Go declaration and asks for a test holding the copies together, and `internal/architecture/native_constants_test.go` already publishes `readNativeSource` as the shared entry point every language-boundary pin uses — including for a Go side behind a build tag the test binary was not built with, which is this case on a non-macOS CI leg. Add the pin using that machinery and add it to ADR 0007's inventory.

The pin should fail on the mistakes that actually happen: a field added to one side, a field reordered, a type widened, or either declaration renamed or moved so the pin reads nothing. Give it its own file rather than extending `cgo_includes_test.go`, which is about include resolution rather than declaration pinning.

## Acceptance criteria

- [ ] A guardrail in `internal/architecture` pins `callback_context.h`'s struct to `overlayutil.CallbackContext` — field names, order, and integer widths
- [ ] It reads both sides through `readNativeSource`, so it runs on every CI leg rather than only on macOS
- [ ] It fails loudly when either declaration is renamed or moved, rather than passing over nothing
- [ ] Verified by adding a field to one side only, and separately by reordering the two fields, and watching each fail
- [ ] The pin is listed in ADR 0007's inventory
- [ ] The comment on `callback_context.h` points at the pin instead of standing in for it
- [ ] `just ci` passes
- [ ] If the other two tickets in this set (#1405, #1406, #1407) are already closed when this lands, flip ADR 0009 from `proposed` to `accepted` in `docs/adr/0009-a-bridge-interface-is-its-headers.md`; if either is still open, leave the status untouched — the flip belongs to whichever of the three lands last

## Blocked by

- None — can start immediately.

y3owk1n:
## Description

This PR adds a guardrail that keeps the macOS overlay's async resize callback
working. The completion path allocates a small record on the C heap, native
code holds it across two dispatch hops, and the Go side reads it back by
casting the raw pointer to its own struct. The two declarations of that record
live in packages that cannot see each other, and until now only a comment said
they had to match — so a field added, reordered, renamed or widened on one side
would have compiled cleanly and then been read from the wrong offsets, ending
as a resize completion silently dropped or dispatched against the wrong
operation.

The guardrail holds the two declarations to the same field names, the same
order and the same integer widths, and fails loudly if either one is renamed or
moved rather than quietly passing over nothing. It also keeps itself honest:
one test applies every way the two could drift and requires the pin to catch
each, and another rewrites the C declaration into shapes the pin does not read
and requires it to report them.

The deliberate limitation is that the C side is read as text, so it recognises
one spelling of the declaration. Rewriting the struct in an equivalent shape
fails the pin rather than being understood — the failure names the shape it
expected, which is the trade every language-boundary pin in this repo makes.

## Related Issues

Closes #1407

## Target Platform

- [x] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — N/A, no build-tagged file changes; the only platform edit is a comment in an existing header.
- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no port or capability changes.

## General Checklist

## Additional Context

**One acceptance criterion was answered differently, on purpose.** The issue
asks for both sides to be read as text "so it runs on every CI leg rather than
only on macOS", on the reasoning that the Go side sits behind a build tag. It
does not — that declaration is untagged and links on every leg. So the C header
is read as text through the shared reader every language-boundary pin here
already uses, and the Go side is reflected over instead. That satisfies the
intent (the pin runs on all three legs, verified by vetting the package for
linux and windows) and is stricter about the two things the issue actually asks
for: reflection reports the real field order and the real width, and a renamed
or moved Go declaration becomes a compile error rather than a skipped read.

**What the pin adds per platform, stated in the test itself.** On macOS the
cgo allocator already names the struct and both members, so a rename or a
changed width is a compile error today — each of those three edits was made and
watched to fail. The reorder is not caught there: both members are the same
width and cgo assigns them by name. Off macOS no cgo is built at all. The
reorder, and every case on the other two legs, is what this pin is the only
reader of.

**ADR bookkeeping.** The pin joins ADR 0007's inventory of pinned
language-boundary copies as its seventh entry. ADR 0009 moves from proposed to
accepted: this was the last of its three tickets, with #1405 and #1406 already
closed and merged.

### DIFF
--- a/docs/adr/0007-a-shared-derivation-has-one-implementation.md
+++ b/docs/adr/0007-a-shared-derivation-has-one-implementation.md
@@ -102,8 +102,8 @@ are the precedent, not the exception.
   predicate. The pin now runs the shared Go rule and reads only the Objective-C
   one, exactly as the label-autohide pin beside it does.
 - **The exception is the half of this rule deletion cannot enforce, so what is
-  pinned is inventoried here.** Six language-boundary copies are pinned as of
-  #1402: the three named above, plus the synthetic key-up and modifier-toggle
+  pinned is inventoried here.** Seven language-boundary copies are pinned as of
+  #1407: the three named above, plus the synthetic key-up and modifier-toggle
   wire prefixes, which `internal/adapter/platform/darwin/eventtap_darwin.m` and
   `internal/adapter/platform/linux/overlay_wayland.c` format with `printf`
   because neither can import `internal/domain/keyvocab`, held to that package by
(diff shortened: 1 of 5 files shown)


y3owk1n:
Good catch on the parser hardening — fixed in b635015c, though not the way the prompt suggested.

Distinguishing an active declaration from a commented-out or `#if 0`-ed one properly means running the preprocessor, which is more machinery than this pin should carry (and would not run off macOS). What the pin can do is refuse to choose: it now reads *every* match of the typedef and reports a header that spells it more than once, rather than silently taking the first. That closes the case that actually happens — an old copy left commented out beside a new one — and does it loudly.

The remaining gap is stated in the code: a header carrying *only* a commented-out copy still parses. That one cgo already catches, since the allocator names the type by hand. There is a new case in `TestCallbackContextLayoutPinReportsADeclarationItCannotRead` covering the duplicate.

PR #1418, opened 2026-08-09
feat(eventtap): let a mode bind the physical mouse buttons

### LINKED ISSUE #1417: Let a mode react to a click from a real mouse
### Category

Configuration

### Target platform

macOS

### Problem or use case

I drive Neru alongside a keyboard that sends a real mouse click from a dedicated key (ZSA Voyager, but any pointing device or mouse-button key behaves the same). When I use recursive grid to place the cursor and then click with that key, the click lands correctly but the grid overlay stays up, and I have to press Escape afterwards to dismiss it.

Nothing in Neru can react to that click today. The mode event tap watches keyboard events only, so a physical button press is invisible to the active mode — there is no binding, flag, or setting that can dismiss an overlay when the user clicks with a real mouse.

The `--action left_click` flag covers the case where Neru performs the click itself, but not the case where the click comes from hardware that Neru does not drive.

### Proposed solution

Report the physical mouse buttons to the active mode as bindable named keys — `MouseLeft`, `MouseRight`, `MouseMiddle` — so they can be written in a mode's hotkey table like any other key:

```toml
[recursive_grid.hotkeys]
"MouseLeft" = "idle"    # clicking anywhere closes the grid
```

Two properties seem necessary for this to be safe:

- **Observe, never consume.** The binding runs in addition to the click reaching the application underneath, so the click still does what it normally does.
- **Ignore Neru's own clicks.** Clicks that Neru performs (`action left_click`, `--action left_click`) must not trigger these bindings, or a binding could recurse into itself.

Being bindable rather than a boolean setting means it composes with macros and works in any mode, and it adds no new config option — which seemed preferable given that every option is a cost.

Since a global hotkey is resolved to a virtual key code and a mouse button has none, a mouse button written in the global `[hotkeys]` table would parse and then never fire; refusing it there with a message pointing at the mode tables avoids that silent failure.

### Alternatives considered

- **A boolean like `recursive_grid.exit_on_external_click`.** Simpler to describe, but it adds a config option, only answers the one question, and cannot express right-click or middle-click, or any action other than exiting.
- **Reprogramming the keyboard firmware** to tap an unused key (F13) before clicking, and binding that to `idle`. Works without any Neru change, but the extra keystroke is then sent everywhere, including outside Neru, so it has to stay unbound in every other application.
- **Doing nothing and using `--action left_click`.** Only helps when Neru performs the click; it cannot see a click from external hardware.

### Related but distinct

The existing mouse work all runs the other direction — Neru *producing* a click. #1102 added right/middle click and the mouse down/up/toggle bindings, and #640 asked for an action that also exits the mode, which shipped as `--action` exiting on completion. Both are about a click Neru performs. This asks for the input direction: a click Neru did not perform reaching the active mode.

### Contribution

Yes — I have this working locally against `main` and will open a PR shortly. It touches the macOS event tap, the shared key vocabulary, and the config reference; `just ci` passes.

mitchellngoodman:
## Description

This PR adds the physical mouse buttons to the key vocabulary, so a mode can bind them the way it binds a key. Until now the mode event tap watched keyboard events only, so a click from a real mouse — or from a keyboard key that sends one — was invisible to the active mode: the click landed, the overlay stayed up, and dismissing it took a second keystroke.

A mode now observes a button press without consuming it, so the binding runs *in addition to* the click reaching the application underneath. Clicks Neru performs itself are marked and ignored, so a binding cannot recurse into its own action. Modifiers are deliberately not encoded into the name — a modified click reports the button alone — because a mode that wants to react to a click wants to react however the click was modified.

Only the macOS tap reports these today; the buttons stay in the shared vocabulary the way `Insert` and F21–F24 do, so one config file still validates on every platform.

## Related Issues

Closes #1417

## Target Platform

- [x] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — the new code sits in existing `_darwin.m` sources, which carry no Go build tag of their own
- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, nothing new is added to a port; the Linux and Windows taps simply never emit these names, the same shape as `Insert` and F21–F24

## General Checklist

## Config surface

Three new named keys, valid in any mode's hotkey table:

| Key | Fires on |
| --- | --- |
| `MouseLeft` | physical left button press |
| `MouseRight` | physical right button press |
| `MouseMiddle` | physical middle button press (button 2 only) |

```toml
[recursive_grid.hotkeys]
"MouseLeft" = "idle"    # clicking anywhere closes the grid
```

They are reported only while a mode is active, so they belong in a mode table. Writing one in the global `[hotkeys]` table is **refused with an error** naming the mode tables instead — a global hotkey resolves to a virtual key code and a mouse button has none, so accepting it would produce a binding that parses and never fires.

No existing option changes, no default changes, and every configuration that loads today still loads: these spellings were already refused as an invalid key format before the vocabulary carried them, so the global-table refusal preserves the behavior they already had.

## Additional Context

**Why bindable keys rather than a setting.** A boolean such as `exit_on_external_click` would answer only the one question, could not express right- or middle-click, and would add a config option. Making the buttons bindable adds none, composes with macros, and works in every mode.

**Telling a physical click from Neru's own.** The tap already ignores events Neru posts, by a marker the keyboard feed sets. The mouse path was not setting it; it is now set in the single helper every synthetic mouse event is posted through, which is what keeps `action left_click` from firing a `MouseLeft` binding.

**Middle button.** Only button number 2 is named `MouseMiddle`; buttons 3 and up stay unbindable rather than being folded onto the middle button.

**Why macOS only.** Not a platform limitation, just a different amount of work, so this PR stays scoped to the platform it can verify. The macOS tap is session-wide and was already receiving these events and discarding them by mask, so widening the mask was the change. The evdev capture is device-scoped to keyboards — a device qualifies by reporting Q/W/E/R/Space/Enter, which a mouse does not — so buttons would need a new device class to enumerate and hotplug-track, and a non-grabbing read path, since the existing capture is an exclusive `EVIOCGRAB` and a button must not be consumed. X11 would need raw XInput2 button events on top of that, and Windows a second low-level hook beside the keyboard one. Happy to follow up on any of these if they are wanted.

**Testing.** `just ci` passes locally, and so do the containerized cross-platform runs: `just test-linux` (CGO on), `just test-linux-nocgo`, `just lint-cross` and `just test-windows-compile`.

The Linux run is what caught the one real cross-platform consequence. `TestX11KeysymFor_EveryNamedKeyResolves` asserts every named key resolves to an X11 keysym, and a mouse button has none — the three new keys failed it. That test is tagged `linux && cgo`, so a macOS-only `just ci` does not compile it. Mouse buttons are now skipped there with the reason recorded: the invariant is about a key a grab could reject, and a button never reaches a grab. Verified in both directions — the three fail without the skip and pass with it.

Beyond the suites, the three buttons were exercised by hand on macOS 15.7.4 against recursive grid: each dismissed the overlay while the click still reached the application underneath.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -345,8 +345,32 @@ which of the two applies to your session.
 | Named      | `Space`, `Return`, `Enter`, `Escape`, `Tab`, `Delete`, `Backspace`                                      |
 | Navigation | `Up`, `Down`, `Left`, `Right`, `Home`, `End`, `PageUp`, `PageDown`, `Insert` (Linux and Windows only)   |
 | Function   | `F1`–`F24` (`F21`–`F24` on Linux and Windows only)                                                      |
+| Mouse      | `MouseLeft`, `MouseRight`, `MouseMiddle` (macOS only, mode tables only — see below)                     |
 
 See [CLI.md](CLI.md#neru-action-feed) for a full key reference with key codes and platform behavior.
+
+**Mouse buttons** are reported only while a mode is active, so they belong in a
+per-mode hotkey table. Writing one in the global `[hotkeys]` table is refused
+with an error rather than accepted and silently never fired. They also behave
+differently from keys in two ways:
+
+- **The click is never swallowed.** The binding runs *in addition to* the click
+  reaching the application underneath, so the click still does whatever it
+  normally would.
+- **Modifiers are not encoded.** `Cmd`-click and `Shift`-click both report
+  `MouseLeft`; there is no `Cmd+MouseLeft`.
+
+Clicks that Neru itself performs (`action left_click`, `--action left_click`) do
+not trigger these, so a binding cannot recurse into itself.
+
+The main use is dismissing an overlay when you click with a real mouse — handy
+if a pointing device or a keyboard with a mouse-button key sits alongside
+keyboard navigation:
+
+```toml
+[recursive_grid.hotkeys]
+"MouseLeft" = "idle"    # Clicking anywhere closes the grid
+```
 
 Multi-key sequences (e.g. `gg`, `ab`) are supported for per-mode hotkeys with a 500ms timeout.
 
--- a/internal/adapter/platform/darwin/accessibility_mouse_darwin.m
+++ b/internal/adapter/platform/darwin/accessibility_mouse_darwin.m
@@ -32,11 +32,21 @@
 /// last leaves the cursor and the viewport agreeing.
 static os_unfair_lock mouseMoveLock = OS_UNFAIR_LOCK_INIT;
 
+/// Marks events Neru posts itself, so the mode event tap can tell its own
+/// synthetic clicks apart from ones the user physically produced. Must match
+/// the marker in keyfeed_darwin.m and the check in eventTapCallback.
+static const int neruSyntheticMouseEventMarker = 0x1337;
+
 /// Pan the zoom viewport to reveal a point and post an event positioned there
 /// @param event Event to post; its location must already be set to position
 /// @param position Position the event acts at
 static void postPositionedEventLocked(CGEventRef event, CGPoint position) {
 	os_unfair_lock_lock(&mouseMoveLock);
+
+	// Tag before posting so the event tap ignores it. Every synthetic mouse
+	// event Neru emits funnels through here, so this is the single point that
+	// keeps `action left_click` from being mistaken for a physical click.
+	CGEventSetIntegerValueField(event, kCGEventSourceUserData, neruSyntheticMouseEventMarker);
 
 	NeruEnsureZoomViewportContainsPoint(position);
 	CGEventPost(kNeruMouseEventTapLocation, event);
--- a/internal/adapter/platform/darwin/eventtap_darwin.m
+++ b/internal/adapter/platform/darwin/eventtap_darwin.m
@@ -588,6 +588,37 @@ CGEventRef eventTapCallback(CGEventTapProxy proxy, CGEventType type, CGEventRef 
 			return event;
 		}
 
+		// Physical mouse buttons are reported to the active mode as the named keys
+		// "MouseLeft", "MouseRight" and "MouseMiddle", so they can be bound in a
+		// mode's [<mode>.hotkeys] table like any other key.
+		//
+		// Two deliberate differences from key handling:
+		//   - The event is always returned, never swallowed. A mode reacting to a
+		//     click must not stop that click from reaching the app underneath.
+		//   - Modifiers are not encoded into the name. Cmd-click and Shift-click are
+		//     ordinary system gestures, and a mode that wants to exit on a click
+		//     wants to do so however the click was modified.
+		//
+		// Events Neru posted itself already returned above via the 0x1337 marker
+		// check, so `action left_click` cannot trigger these.
+		if (type == kCGEventLeftMouseDown || type == kCGEventRightMouseDown || type == kCGEventOtherMouseDown) {
+			NSString *buttonName = nil;
+			if (type == kCGEventLeftMouseDown) {
+				buttonName = @"MouseLeft";
+			} else if (type == kCGEventRightMouseDown) {
+				buttonName = @"MouseRight";
+			} else if (CGEventGetIntegerValueField(event, kCGMouseEventButtonNumber) == 2) {
+				// Only button 2 is the middle button; buttons 3+ stay unbindable.
+				buttonName = @"MouseMiddle";
+			}
+
+			if (buttonName && context->callback) {
+				context->callback([buttonName UTF8String], context->userData);
+			}
+
+			return event;
+		}
+
 		if (type == kCGEventKeyUp) {
 			CGKeyCode keyCode = (CGKeyCode)CGEventGetIntegerValueField(event, kCGKeyboardEventKeycode);
 			NSString *keyName = specialKeyName(keyCode);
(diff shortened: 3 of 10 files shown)


y3owk1n:
i will comment on this later when i am done and happy with my first round of refactoring the codebase, so that you don't have to rebase that often.

internal/adapter/platform/darwin/eventtap_darwin.m
```diff
@@ -588,6 +588,37 @@ CGEventRef eventTapCallback(CGEventTapProxy proxy, CGEventType type, CGEventRef
...
+		// Two deliberate differences from key handling:
+		//   - The event is always returned, never swallowed. A mode reacting to a
+		//     click must not stop that click from reaching the app underneath.
+		//   - Modifiers are not encoded into the name. Cmd-click and Shift-click are
+		//     ordinary system gestures, and a mode that wants to exit on a click
+		//     wants to do so however the click was modified.
+		//
+		// Events Neru posted itself already returned above via the 0x1337 marker
+		// check, so `action left_click` cannot trigger these.
+		if (type == kCGEventLeftMouseDown || type == kCGEventRightMouseDown || type == kCGEventOtherMouseDown) {
```
y3owk1n, line 604: I think we have to also consider what happens when `held_repeat.enabled` is enabled. Looking at it, I think it will repeat the run forever until the mode is idle, probably not something expected. Maybe a guard in `key_dispatch.go` is needed around the repeat?

internal/config/validate.go
```diff
@@ -192,6 +193,22 @@ func (c *Config) ValidateHotkeyBindings() error {
 			return err
 		}
 
+		// A global hotkey is resolved to a virtual key code, which no mouse
+		// button has. Left to validate, the binding would parse and then never
+		// fire — the silent failure a mode binding exists to avoid. Refusing
+		// keeps the behavior these keys already had before the vocabulary
+		// carried them, so no configuration that loads today stops loading.
+		if keyvocab.IsMouseButton(key) {
```
y3owk1n, line 201: Looks like this test the whole key, i think the validation might get skipped if a user binded it as `Cmd+MouseLeft` or some sort. Worth a double check.

internal/adapter/platform/darwin/eventtap_darwin.m
```diff
@@ -588,6 +588,37 @@ CGEventRef eventTapCallback(CGEventTapProxy proxy, CGEventType type, CGEventRef
...
+			NSString *buttonName = nil;
+			if (type == kCGEventLeftMouseDown) {
+				buttonName = @"MouseLeft";
```
y3owk1n, line 607: Referring to the literals of mouse left, right and middle, i think we should add them into the keyvocab `nativeKeyEventEmitters` for consistency, as that's part of the guardrail to prevent weird cases in the future.

internal/adapter/platform/darwin/accessibility_mouse_darwin.m
```diff
@@ -32,12 +32,22 @@
 /// last leaves the cursor and the viewport agreeing.
 static os_unfair_lock mouseMoveLock = OS_UNFAIR_LOCK_INIT;
 
+/// Marks events Neru posts itself, so the mode event tap can tell its own
+/// synthetic clicks apart from ones the user physically produced. Must match
+/// the marker in keyfeed_darwin.m and the check in eventTapCallback.
+static const int neruSyntheticMouseEventMarker = 0x1337;
```
y3owk1n, line 38: i think moving this to `accessibility_constants.h` probably make more sense.

internal/domain/keyvocab/named_keys.go
```diff
@@ -29,8 +29,43 @@ const (
...
+// mouseButtons is the subset of the vocabulary that names a mouse button
+// rather than a keystroke. It exists so the global-hotkey validator can refuse
+// one: a global hotkey is resolved to a virtual keycode, which no mouse button
+// has, so a binding written there would parse and then never fire.
+var mouseButtons = []string{KeyMouseLeft, KeyMouseRight, KeyMouseMiddle}
+
+// IsMouseButton reports whether name is one of the mouse buttons,
+// case-insensitively.
+func IsMouseButton(name string) bool {
+	lowered := strings.ToLower(name)
```
y3owk1n, line 53: we already have lowercase helper down below, we probably can just use them for consistency.

internal/domain/keyvocab/named_keys.go
```diff
@@ -29,8 +29,43 @@ const (
...
+	return slices.ContainsFunc(mouseButtons, func(button string) bool {
+		return strings.ToLower(button) == lowered
+	})
+}
+
+// MouseButtons returns every mouse button in display form, sorted, so a caller
+// listing them in a diagnostic reads the declaration rather than restating it.
+func MouseButtons() []string {
+	buttons := slices.Clone(mouseButtons)
+	slices.Sort(buttons)
```
y3owk1n, line 64: i think sorting this will fight against the comment where it should be `reads the declaration` no? We will need to pick between either keep sort and rewrite the comments, or another way round.

docs/CONFIGURATION.md
```diff
@@ -345,9 +345,33 @@ which of the two applies to your session.
...
+**Mouse buttons** are reported only while a mode is active, so they belong in a
+per-mode hotkey table. Writing one in the global `[hotkeys]` table is refused
+with an error rather than accepted and silently never fired. They also behave
+differently from keys in two ways:
+
+- **The click is never swallowed.** The binding runs *in addition to* the click
+  reaching the application underneath, so the click still does whatever it
+  normally would.
+- **Modifiers are not encoded.** `Cmd`-click and `Shift`-click both report
+  `MouseLeft`; there is no `Cmd+MouseLeft`.
```
y3owk1n, line 361: `there is no Cmd+MouseLeft`, this is true, but we got no validation to reject this also, i think a rephrase is needed.

docs/CONFIGURATION.md
```diff
@@ -345,9 +345,33 @@ which of the two applies to your session.
 | Named      | `Space`, `Return`, `Enter`, `Escape`, `Tab`, `Delete`, `Backspace`                                      |
 | Navigation | `Up`, `Down`, `Left`, `Right`, `Home`, `End`, `PageUp`, `PageDown`, `Insert` (Linux and Windows only)   |
 | Function   | `F1`–`F24` (`F21`–`F24` on Linux and Windows only)                                                      |
+| Mouse      | `MouseLeft`, `MouseRight`, `MouseMiddle` (macOS only, mode tables only — see below)                     |
 
 See [CLI.md](CLI.md#neru-action-feed) for a full key reference with key codes and platform behavior.
 
+**Mouse buttons** are reported only while a mode is active, so they belong in a
```
y3owk1n, line 352: Since we have some info about the refusal, i think it's worth to tell the user that the eventtap (mouse clicks in this case) will be ignored when entered the `search input` mode, probably not alot of users using that, but worth a message i think.

y3owk1n (changes requested):

PR #1444, opened 2026-08-10
build(justfile): give every recipe an accurate summary in just --list

### LINKED ISSUE #1439: build(justfile): declare every recipe's listed documentation
## What to build

`just --list` is wrong for **26 of the 52** public recipes, and has been for months. The cause is not a typo in any comment — every comment block is accurate prose. It is that `just` reads only the *last contiguous comment line* before a recipe, and this justfile writes multi-line rationale paragraphs there. So the listed text is the tail of a paragraph:

- `build` → "*X11/Wayland native backends). Windows currently builds with CGO disabled.*"
- `list-foundation-packages` → "*check could see.*"
- `test` → "*test-desktop\` to include them when you can hand the machine over.*" — opening on an unmatched backtick
- `genflagref` → "*guardrail fails while the page is out of date.*"
- `vuln` → "*vulnerability database is fetched at run time regardless.*"

`docs/DEVELOPMENT.md:107` names `just --list` under *Verify* as the way a newcomer sees what the project can do, so this is the orient step of the first hour reading as noise.

The layout fix does not hold. `build` already tried it — a blank line mid-block — and still renders wrong, because its summary sits on the far side of the blank. Which line `just` happens to read is a function of where someone put whitespace, so it re-breaks silently the next time the rationale is edited.

Declare it instead. `just 1.54.0` (pinned in `devbox.json`) supports the `[doc('…')]` attribute, which overrides the comment entirely and leaves the prose comments exactly where they are. Verified: `just --dump --dump-format json` then carries the declared text per recipe.

This is [ADR 0012](../blob/main/docs/adr/0012-the-first-hour-must-not-lie.md) — *the first hour must not lie* — and the *Option* entry in `CONTEXT.md` applied to the build system: one declaration, and the rendered list is a projection of it.

It also earns a guardrail by ADR 0011's criterion. The justfile parses, every recipe runs, `just lint` is silent and every test passes with the list wrong — which is exactly how it stayed wrong.

## Acceptance criteria

- [ ] Every public recipe in `justfile` carries a `[doc('…')]` attribute whose text is a one-line summary of what the recipe does — not the tail of its rationale.
- [ ] The existing prose comment blocks are left where they are, unedited. This ticket does not rewrite rationale.
- [ ] `just --list` reads as 52 summaries with no sentence fragments and no unmatched backticks.
- [ ] `internal/architecture/justfile_doc_test.go` parses `justfile` **as text** — no shelling out to `just`, per ADR 0011's note that `internal/architecture` reads the tree as text so it holds on every CI leg — and fails when a public recipe has no `[doc('…')]` attribute.
- [ ] The guardrail also fails on an empty doc string and on one long enough to wrap in `just --list` at 80 columns.
- [ ] The guardrail carries a floor assertion so the walk cannot silently match zero recipes, per the house pattern.
- [ ] Its failure message states the rule in the imperative and cites ADR 0012. No `AGENTS.md` sentence cites the test back.
- [ ] `internal/architecture/doc.go` gains its inventory entry (`doc_inventory_test.go` holds this in both directions).

## Blocked by

Nothing. Concurrent with the `-v` split and `check-cross` tickets — all three touch different lines of `justfile`.

y3owk1n:
## Description

This PR fixes `just --list`, which described 26 of the 53 recipes with the tail
of a sentence: `build` advertised itself as "*X11/Wayland native backends).
Windows currently builds with CGO disabled.*", `list-foundation-packages` as
"*check could see.*", and `test` as a fragment opening on an unmatched
backtick. Nothing was mistyped — `just` reads only the last contiguous comment
line before a recipe, and this justfile writes multi-line rationale there, so
the listed text was whichever line a blank happened to leave last.

Every recipe now declares its own one-line summary with a `[doc('…')]`
attribute, which overrides the comment block entirely. The rationale comments
are untouched and stay exactly where they were, so `just --list` reads as 53
summaries and the listed text no longer moves when someone edits the prose
above it.

A guardrail in `internal/architecture` reads the justfile as text — it never
shells out to `just` — and fails when a recipe `just --list` shows declares no
summary, declares an empty one, or declares a paragraph. One deliberate
limitation: the width rule caps the summary text rather than the rendered list
line, for the reason in Additional Context.

## Related Issues

Closes #1439

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix
- [x] `chore` — Build, CI, dependencies, tooling

## Cross-Platform Checklist

- [ ] No darwin imports from untagged (shared) code — [The One Rule](https://github.com/y3owk1n/neru/blob/main/docs/ARCHITECTURE.md#the-one-rule)

## General Checklist

      check, lint, vet, tests including a unit `-race` pass, vulnerability
      scan, build); CI runs them on macOS, Linux and Windows
- [x] Documentation updated (if applicable) — the recipe summaries are the
      documentation this changes, and no markdown page quotes `just --list`
      output

## Additional Context

**Why a guardrail.** ADR 0012 decides that recipe documentation is declared
rather than inferred from layout, and that it earns a guardrail by ADR 0011's
criterion: the justfile parses, every recipe runs, `just lint` is silent and
every test passes with the list wrong, which is how it stayed wrong for months.
The new test was mutation-checked against the real justfile — deleting a
summary, emptying one, and inflating one each fail with their own message.

**The one judgement call: what "too long" means.** The issue asks the guardrail
to fail on a summary "long enough to wrap in `just --list` at 80 columns".
Taken as the rendered line, that is unreachable. `just` pads every signature to
the widest one it will align — `generate-icons APP_ICON TRAY_ACTIVE
TRAY_DISABLED`, 49 columns — so the summary column starts at 56 and an
80-column terminal leaves 24 columns of text. Summaries that short are labels,
not sentences, and they would be a downgrade for the 27 entries that already
read correctly today at 40 to 76 columns. So the constant caps the summary text
at 80, with the arithmetic written out beside it, and the honest consequence is
that every rendered line still wraps on an 80-column terminal (they run 105 to
128 columns). If you would rather the rule bite on wrapping, a rendered-line
budget of around 120 columns is a one-line change plus a reflow of the four
longest summaries — that number is yours to pick, not mine.

**Recipe count.** The issue and ADR say 52 recipes; there are 53 today, all
public, and all 53 carry a summary.

**Shared parser.** The justfile parser `foundation_slice_test.go` already owned
now also records the attributes above each recipe, and is split so a fixture
can be parsed as text. There is deliberately no second justfile parser in the
package — though the parser arguably now wants a file of its own, the way the
shared repo walk and the shared native-constant reader have one.

### DIFF
--- a/internal/architecture/doc.go
+++ b/internal/architecture/doc.go
@@ -39,6 +39,9 @@
 //     by naming it, and every test name or test file it names resolves.
 //   - hint_placement_vocabulary_test.go — the hint placement vocabulary is the
 //     same on both sides of the Go/Objective-C boundary.
+//   - justfile_doc_test.go — every recipe just --list shows declares its own
+//     one-line summary, rather than inheriting whichever line of the comment
+//     block above it just happens to read.
 //   - keyvocab_wire_test.go — the native key event emitters and keyvocab agree
 //     on what they put on the wire.
 //   - label_autohide_rule_test.go — the label autohide rule is pinned across
--- a/internal/architecture/foundation_slice_test.go
+++ b/internal/architecture/foundation_slice_test.go
@@ -391,12 +391,16 @@ func recipePackages(t *testing.T, recipes justRecipes) []string {
 // justRecipes is the justfile read as a graph, keyed by recipe name.
 type justRecipes map[string]justRecipe
 
-// justRecipe is one recipe: the indented lines under its header, and the
-// recipes running it runs — its header dependencies plus any `just <name>` call
-// in that body.
+// justRecipe is one recipe: the indented lines under its header, the
+// attributes declared immediately above it, and the recipes running it runs —
+// its header dependencies plus any `just <name>` call in that body.
 type justRecipe struct {
 	body string
 	runs []string
+	// attributes is the `[…]` lines between the recipe's comment block and its
+	// header, joined as they appear. justfile_doc_test.go reads the declared
+	// summary out of them.
+	attributes string
 }
 
 // mustFind returns a recipe, failing the test when the justfile has none by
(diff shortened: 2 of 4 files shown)


y3owk1n:
**Width rule: resolved — keep the 80-character text cap as shipped.**

The acceptance criterion in #1439 asked the guardrail to fail on a summary "long enough to wrap in \`just --list\` at 80 columns". That criterion was unreachable and this PR was right not to satisfy it literally: \`just\` pads the signature column to the widest signature (\`generate-icons APP_ICON TRAY_ACTIVE TRAY_DISABLED\`, 49 chars), so summaries begin at column 56 and an 80-column terminal leaves 24 characters — a label, not a summary, and a downgrade for the 27 entries that read correctly today.

A rendered-width budget of 120 columns was considered and declined. What the guardrail is for is "a summary is a line, not a paragraph", and the text cap states exactly that. Enforcing rendered width would also couple every recipe's pass/fail to the longest signature in the file, so adding one long-signature recipe would fail the guardrail on unrelated lines.

No change requested. Ready for review.

PR #1446, opened 2026-08-10
build(justfile): catch Linux and Windows build breaks in the pre-push gate

### LINKED ISSUE #1441: build(justfile): add check-cross to the pre-push gate
## What to build

`just ci` is `fmt-check lint vet build test-ci vuln` — the right six recipes, on one host. CI runs the same recipes as a 17-job matrix (3 OS × 5 checks, plus `formatting` and `coverage` on macOS), so a macOS contributor can pass the documented pre-push gate and fail ten jobs.

#1438 already fixed the *prose* — `AGENTS.md:41`, `CONTRIBUTING.md:87`, the PR template and the recipe comment now all say "on your host only, where CI runs them on macOS, Linux and Windows". That wording is now the contract. This ticket narrows the gap the contract admits to.

Add `check-cross`: a CGO-off type-check of the Linux and Windows builds. It costs seconds, needs no Docker, and catches the single most common cross-platform failure — a build break in a tagged file that `just lint` cannot see, since golangci-lint honours the host's build tags.

`lint-cross` and `test-linux` are deliberately **not** added, and [ADR 0012](../blob/main/docs/adr/0012-the-first-hour-must-not-lie.md) records why: both require a running Docker daemon, and making the documented pre-push gate fail on missing infrastructure is precisely the first-hour failure this work is about. CI remains the check for the cgo-only Linux paths, as `CROSS_PLATFORM.md:999-1001` already says.

## Acceptance criteria

- [ ] `ci` gains `check-cross` in its dependency list, positioned so a plain build break surfaces before the test suites run.
- [ ] The recipe comment states what the addition buys and that Docker-dependent recipes are deliberately excluded, citing ADR 0012.
- [ ] `just ci` still passes on a clean tree, and the added wall-clock is reported in the PR body.
- [ ] A deliberate build break in a `//go:build linux` file fails `just ci` on a macOS host — verify this, it is the whole point.
- [ ] Any prose that enumerates what `just ci` runs is updated to include it: `AGENTS.md:41`, `CONTRIBUTING.md:87-93`, `.github/pull_request_template.md:42-44`. Check for others rather than trusting this list.
- [ ] `.agents/skills/create-pr/SKILL.md:34` still says `just ci` "is exactly what CI runs" — the last surviving instance of the overclaim #1438 fixed elsewhere. It is also fixed by the documentation ticket filed alongside this one; whichever lands first takes it, and the other drops it rather than conflicting.

## Blocked by

Nothing. Concurrent with the `[doc('…')]` and `-v` tickets — different lines of `justfile`.

y3owk1n:
## Description

This PR makes the documented pre-push gate catch cross-platform build breaks, and stops four documents telling contributors that their branch commit subjects are what reach the changelog.

`just ci` now type-checks the Linux and Windows builds alongside the host one. Until now an undefined symbol in a `//go:build linux` file passed the entire local gate on a macOS host and first appeared as a red CI job on a leg the contributor had no way to run — the lint and vet steps compile for the host, so neither could see it. The new step catches it, costs about fifteen seconds, and needs no Docker. It runs with CGO off, so it covers the plain tagged files and not the cgo-only Linux backends; CI stays the check for those, and every place that describes the step says so.

Separately, five places said Release Please ships the subject you write on the branch. This repository squash-merges and enables no other merge method, so the PR title is what ships and a branch subject never can. The convention itself is unchanged — branch commits are still conventional commits — but the reason given for it is now a true one.

> **This PR contains three unrelated commits.** That is deliberate and agreed with the maintainer, not an oversight:
>
> 1. `build(justfile): catch Linux and Windows build breaks in the pre-push gate`
> 2. `docs: name the squash title as what reaches the changelog`
> 3. `test(architecture): give the shared justfile parser its own home`
>
> Because the merge is a squash, only the title above reaches `main` and the changelog; commits 2 and 3 will not be separately visible there. Reviewing commit by commit is the way to read this one.

## Related Issues

Closes #1441

## Target Platform

- [ ] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix
- [x] `chore` — Build, CI, dependencies, tooling — the title's type is `build`; the other two commits are `docs` and `test`

## Cross-Platform Checklist

- [ ] No darwin imports from untagged (shared) code — [The One Rule](https://github.com/y3owk1n/neru/blob/main/docs/ARCHITECTURE.md#the-one-rule)

## General Checklist

- [x] `just ci` passes — the same checks CI runs, on your host only (format check, lint, vet, build, a CGO-off type-check of the Linux and Windows builds, tests including a unit `-race` pass, vulnerability scan); CI runs them on macOS, Linux and Windows
- [x] Tests added/updated for new or changed functionality — N/A, no behaviour to test: the gate change is a recipe dependency, the doc change is prose, and the parser move keeps every existing assertion where it was
- [x] PR title is a [conventional commit](https://www.conventionalcommits.org/) subject written for users — this PR squash-merges, so the title is what Release Please ships in the changelog, not the commits on the branch

## Commands this changes

No new recipe — `just check-cross` already existed. What changes is that `just ci` now runs it:

```
ci: fmt-check lint vet build check-cross test-ci vuln
```

It sits after `build`, so a plain compile error surfaces before the test suites start. `just lint-cross` and `just test-linux` stay out of the gate on purpose: both need a running Docker daemon, and a documented pre-push gate that fails for want of infrastructure is the first-hour failure [ADR 0012](https://github.com/y3owk1n/neru/blob/main/docs/adr/0012-the-first-hour-must-not-lie.md) is about. Both remain available on demand for a real Linux lint or test run.

## Additional Context

**Verified rather than assumed.** With `var _ = deliberateBuildBreakForVerification` appended to `internal/adapter/platform/linux/font_common.go`, on this macOS host:

| | before | after |
| --- | --- | --- |
| `just vet` | exit 0 | exit 0 |
| `just lint` | exit 0 | exit 0 |
| `just build` | exit 0 | exit 0 |
| `just check-cross` | not run by `just ci` | **exit 1** — `font_common.go:45:9: undefined: deliberateBuildBreakForVerification` |

**Cost.** `just check-cross` alone: **23.9s** on a cold cross-target build cache, **15.5s** warm. The Linux and Windows object caches are separate from the host's, so `vet` and `build` running first do not warm them — budget ~16s of steady-state addition, ~24s the first time. For scale, the full `just ci` on this branch took 4m51s cold and 2m33s warm.

**Squash-title correction.** The repository's settings were read directly rather than inferred from commit shapes: squash merging is enabled, merge commits and rebase merging are not. Five homes were corrected, one more than the issue listed — the extra is the Codex overlay beside the `create-pr` skill. The PR template needed no change; it has said this correctly since #1443.

**One judgement call worth a second opinion.** The issue says the correction is "to the stated *reason*, not the rule". Two edits in the `create-pr` skill do touch a rule, and I made them because the misconception had left the rule unable to do its job: its ban on breaking-change markers said "in the commit", when under squash the *title* is where a `!` would actually trigger a major bump, and its list of changelog-visible types did not say whose type is read. Both now name the title. Revert those two if you would rather keep the rule exactly as it stood.

**Documentation reach, and one home for the contract.** `DEVELOPMENT.md`'s Common Tasks table calls itself "the single reference" for every build, test and lint entry point, and it listed neither `just ci` nor `just check-cross`. Both are now rows in it, and the table is followed by the one authoritative statement of what the gate covers and what it does not. The ownership table in `CROSS_PLATFORM.md` puts build recipes there, so the other documents reference it rather than restating the CGO caveat in their own words — `AGENTS.md` and `CONTRIBUTING.md` each name the step and point at it, and the recipe comment in the justfile keeps its rationale and defers on coverage. `CROSS_PLATFORM.md`'s section on `just lint` seeing only your own platform now says a build break is caught before you push; its hand-rolled `golangci-lint` snippet is unchanged and still correct, since `check-cross` runs `go vet` rather than the linter.

**One acceptance criterion is dropped, not missed.** The issue's last strand-1 criterion expects `.agents/skills/create-pr/SKILL.md` to still claim `just ci` "is exactly what CI runs". #1443 already replaced that line, and the criterion itself provides for whichever ticket lands first taking it.

**Parser move.** The whole reader moves, not only the three symbols the issue named — moving the type without its methods would have left the seam in the wrong place. The fixture-driven tests #1444 added stay in `justfile_doc_test.go`, because what they assert on is the summary rule rather than the parser. `withoutComments` and `readRepoFile` stay in `foundation_slice_test.go` too: both also serve the CI-workflow reader beside them, and neither is specific to reading a justfile. `doc_inventory_test.go` holds the new entry in both directions.

**Deliberate limitation.** Nothing fails if `check-cross` is later dropped from `ci` — the guardrail suite pins that `ci` reaches `test-foundation`, but there is no equivalent pin for this step, so the prose could go back to over-promising silently. Worth a follow-up if the gate's membership is something the repo wants held; out of scope here.


y3owk1n:
Addressed the Greptile finding on `AGENTS.md:41` (check-cross guidance having multiple owners). It was right, and the ownership table in `CROSS_PLATFORM.md` names the home: *a build recipe or test tier → DEVELOPMENT.md*.

`DEVELOPMENT.md` now carries the single statement of what the gate covers and what it does not, under **What `just ci` covers, and what it does not**, plus the two table rows that were missing — its Common Tasks table calls itself "the single reference" for every build, test and lint entry point and listed neither `just ci` nor `just check-cross`. `AGENTS.md` and `CONTRIBUTING.md` are back to naming the step and pointing there, `CROSS_PLATFORM.md` no longer restates the CGO mode, and the justfile comment keeps its rationale but defers on coverage.

Force-pushed as a rebuild so the three commits keep their one-strand-each shape rather than growing a fourth review-fix commit. Gate re-run green.

PR #1467, opened 2026-08-11
fix(cli): open docs on Linux

### LINKED ISSUE #1454: fix(cli): open the docs on Linux, where the tray already knows how
## What to build

`neru docs` shells out to `/usr/bin/open` on macOS (`internal/cli/docs_darwin.go`) and returns `CodeNotSupported` everywhere else (`internal/cli/docs_other.go`). Meanwhile `internal/app/components/systray/open_other.go` in the same repository already opens URLs on Linux through `xdg-open`.

So this is not a missing capability with a design question behind it — the code exists, ten lines away, and one caller was never pointed at it. Give `neru docs` the same treatment on Linux.

Windows keeps its current behavior unless the same trick applies there.

## Acceptance criteria

- [ ] `neru docs` opens the documentation on Linux via `xdg-open`
- [ ] A missing `xdg-open` reports an error naming what is absent, not a silent success
- [ ] The two call sites share one implementation rather than each spawning their own
- [ ] `docs/CROSS_PLATFORM.md` Known Gaps entry 3 is removed
- [ ] `just ci` passes

## Blocked by

Nothing. This is the smallest item in the parity set and a reasonable first contribution.

## Parent

#1466 — Linux parity

nightcityblade:
## Description

Teach `neru docs` to open version-aware documentation URLs with `xdg-open` on Linux. The CLI and systray now share the platform opener, and command failures are wrapped as `CodeExecFailed` while preserving details such as a missing `xdg-open` executable.

## Related Issues

Fixes #1454

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

## General Checklist

- [ ] Tests added/updated for new or changed functionality — N/A: this reuses the existing process-launch boundary; the complete unit, integration, race, and cross-compile suite passes.
- [ ] Documentation updated (if applicable) — N/A: the obsolete CLI-docs Known Gaps item mentioned in the issue is already absent from current `main`; the unrelated GNOME gap remains intact.

## Screenshots / Recordings

N/A — command-line behavior only.

## Additional Context

Windows retains its existing explicit `CodeNotSupported` behavior.


internal/adapter/platform/platform.go
```diff
@@ -46,3 +50,15 @@ func IsLinux() bool {
 func IsWindows() bool {
 	return CurrentOS() == Windows
 }
+
+// OpenExternal opens a URL or file path with the Linux desktop's default handler.
+func OpenExternal(ctx context.Context, target string) error {
```
y3owk1n: i think we probably dont need this in platform, since we already have `systray` and `cli` handlers, lets just do it over there.

internal/adapter/platform/platform.go
```diff
@@ -46,3 +50,15 @@ func IsLinux() bool {
...
+	if !IsLinux() {
+		return derrors.New(
+			derrors.CodeNotSupported,
+			"opening external targets is only implemented for Linux",
+		)
+	}
+
+	return exec.CommandContext(ctx, "xdg-open", target).Run()
```
y3owk1n: When doing `Run` it might block the shell until browser is closed. Either we should take a timeout or just use `Start` instead of `Run` to avoid waiting.

internal/app/components/systray/open_other.go
y3owk1n, line 1: i feel i feel like we should just make `open_other` returns unsupported, and write the implementation in `open_linux` instead via proper build tags.

internal/cli/docs_other.go
y3owk1n, line 1: same as `systray`, i think we should keep `docs_other` as unsupported and implement it in `docs_linux` with build tag.

y3owk1n (changes requested):
Also i think the following docs needs to be updated too:

- docs/CROSS_PLATFORM.md:618
- docs/CLI.md:102
- docs/CLI.md:1528
- internal/cli/docs.go:8-10

internal/adapter/platform/platform.go
nightcityblade: Removed the shared platform helper. Linux launching now stays in the build-tagged systray and CLI handlers.

internal/adapter/platform/platform.go
nightcityblade: Switched both Linux launch paths to Start, so the command returns after launching instead of waiting for the browser to close.

internal/app/components/systray/open_other.go
nightcityblade: Done: open_other.go now returns CodeNotSupported on the remaining platforms, and the xdg-open implementation lives in open_linux.go behind the linux build tag.

internal/cli/docs_other.go
nightcityblade: Done: docs_other.go remains the unsupported stub, and docs_linux.go contains the xdg-open implementation behind the linux build tag.

nightcityblade:
Addressed the review in bcce8ca and rebased onto current `main`:

- moved Linux launching into `open_linux.go` / `docs_linux.go` build-tagged handlers
- restored explicit unsupported stubs for other platforms
- switched Linux launches from blocking `Run` to `Start`
- updated the cross-platform gap list, CLI platform table/section, and command documentation

The full `just ci` gate passes, including lint, host/cross vet and build checks, foundation/unit/race/integration tests, and vulnerability scan.

y3owk1n (dismissed):

y3owk1n:
Everything looks good to me now, good to merge after CI turns green.

y3owk1n (approved):
reapprove after formatting

PR #1474, opened 2026-08-11
fix(linux): make modified scroll work on hyprland

DarellYuhu:
## Description

This PR fixes modified scroll actions on Hyprland.

The upstream Wayland implementation routes modified scrolling entirely through
its virtual-pointer path. On Hyprland, that path can fail to produce a scroll
event. Modified scrolling now keeps the reliable uinput scroll path, waits
briefly for the injected modifier state to reach the compositor, and releases
the modifier after scrolling. Unmodified scrolling remains unchanged.

## Related Issues

None.

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags
- [x] Stub implementations added for other platforms

## General Checklist

- [x] `just ci` passes — the macOS Objective-C lint step is skipped on Linux because the macOS SDK is unavailable
- [x] PR title is a [conventional commit](https://www.conventionalcommits.org/) subject written for users

## Additional Context

The modifier flag remains available as `--modifier`, including comma-separated
values such as `ctrl,alt`. The change preserves the existing upstream modifier
handling while selecting the input path that works reliably with Hyprland.
A short synchronization delay is used because the virtual keyboard and uinput
scroll events arrive through separate input sources.


y3owk1n:
@DarellYuhu will you be able to rebase main first? Sorry that i think in the mid i had done some changes.

DarellYuhu:
@y3owk1n I’ve rebased the branch onto `main` and made some adjustments related to the smooth-scrolling feature introduced in the main branch.

internal/adapter/accessibility/native/linux/element.go
```diff
@@ -480,13 +484,17 @@ func scrollAtCursorNow(deltaX, deltaY int, modifiers action.Modifiers) error {
 	}
 
 	if currentLinuxBackend() == linuxBackendWayland {
```
y3owk1n, line 497: i think this hits every wayland compositor and not just hyprland, `currentLinuxBackend()` collapses wlroots and KDE into one `linuxBackendWayland`. So on KWin the modifier goes out on libei and the scroll on uinput. Probably worth gating on the hyprland detection instead.

internal/adapter/accessibility/native/linux/element.go
```diff
@@ -480,13 +484,17 @@ func scrollAtCursorNow(deltaX, deltaY int, modifiers action.Modifiers) error {
...
+		err := pressWaylandScrollModifiers(modifiers)
+		if err != nil {
+			return err
+		}
+		defer func() { _ = releaseWaylandScrollModifiers(modifiers) }()
```
y3owk1n: We sleep before the scroll but not before the deferred release, and the batch write only queues into the kernel. i think the modifier can go up while the scroll is still being processed, worth a double check.

internal/adapter/accessibility/native/linux/element.go
```diff
@@ -480,13 +484,17 @@ func scrollAtCursorNow(deltaX, deltaY int, modifiers action.Modifiers) error {
...
+
+		// Keyboard and uinput events arrive through separate input sources.
+		// Let Hyprland apply the modifier before the scroll device reports it.
 		if modifiers != 0 {
```
y3owk1n, line 507: 5ms is a guess and nothing confirms the modifier landed. Under load it'd scroll unmodified and still report success — maybe we can wait on something real instead?

internal/adapter/accessibility/native/linux/element.go
```diff
@@ -428,21 +430,23 @@ const scrollPixelsPerNotch = 30
...
-// because the animation runs on a worker goroutine where a refusal has nobody
-// to go back to — and a refusal that reaches nobody is exactly the silent
-// no-op turning this option on must not introduce.
+// arrives as a sequence of eased chunks. Modified Wayland scrolls use the
+// immediate uinput path instead, because Hyprland does not reliably merge the
+// virtual keyboard modifier with virtual-pointer scroll events.
 func ScrollAtCursor(deltaX, deltaY int, modifiers action.Modifiers) error {
 	cfg := currentLinuxConfig()
 	if cfg != nil && cfg.SmoothScroll.Enabled {
+		if currentLinuxBackend() == linuxBackendWayland && modifiers != 0 {
```
y3owk1n: This returns above `scrollBackendAvailable()` and the zero-delta guard below, and `scrollAtCursorNow` has no zero-delta check of its own, so an empty modified scroll still presses and releases a real key. It also means `smooth_scroll` silently stops applying to modified scrolls.

internal/adapter/accessibility/native/linux/element.go
```diff
@@ -570,6 +578,54 @@ func scrollAtCursorNow(deltaX, deltaY int, modifiers action.Modifiers) error {
 	return nil
 }
 
+func pressWaylandScrollModifiers(modifiers action.Modifiers) error {
```
y3owk1n: i think the unwind on error releases modifiers we never pressed, and the dispatcher sends a key-up even when the count is already 0, so it could clear one the user is physically holding. Probably worth only unwinding what we actually pressed.

internal/adapter/accessibility/native/linux/element.go
y3owk1n, line 598: Now that the modified path is gone, this is the only caller left and it passes 0, so the `modifiers` param on `wlrootsScrollAtCursor` and its comment are dead. Not this PR's job, but worth a follow-up.

y3owk1n, line 1: I think we also need to update the matrix row and the prose around line 428 for `cross_platform.md`, where both still say the uinput batch is skipped, which is the opposite of this change now.

y3owk1n (changes requested):

y3owk1n:
@DarellYuhu i just left some comments, since linux is a bit complicated with different compositors or even variants, we have to ensure we covers them properly.

y3owk1n:
I've merged latest main into here and fixes some stuffs to be more specific to hyprland.

y3owk1n (approved):

y3owk1n:
merging this and big thanks! let me know if my fix breaks what your expectation, and we can fix them later again.

DarellYuhu:
Build from latest main branch and it works. Thanks 🎉

PR #1502, opened 2026-08-12
refactor(linux): share one shm-file helper across the Wayland backends

### LINKED ISSUE #1479: refactor(linux): one shm-file helper for the three Wayland call sites
## What to build

Three C files each carry their own private "make me an anonymous fd of N bytes" helper, and they have already drifted:

- `internal/adapter/platform/linux/overlay_wayland.c:66` — `create_shm_file(off_t)`: `memfd_create` behind `#ifdef __NR_memfd_create`, with a `mkstemp` + `unlink` fallback
- `internal/adapter/platform/linux/wlroots_client.c:271` — `neru_wlr_create_shm_file(off_t)`: the same shape, same fallback
- `internal/adapter/platform/linux/wlroots_screencopy.c:355` — `neru_screencopy_shm_file(size_t)`: added by #1459, **`memfd_create` only** — no `#ifdef` guard and no `mkstemp` fallback

The drift is the point. The newest copy is the least portable of the three: on a target without `__NR_memfd_create` the other two still produce a buffer and screencopy does not. Nobody decided that; it is what copying a helper by hand costs on the third repetition. A fourth backend is coming with the KDE portal path (#1478).

Consolidate into one shared helper in its own `.c`/`.h` pair alongside the other shared Linux C, and have all three call it. `neru_wlr_create_keymap_fd` (`wlroots_client.c:393`) is deliberately **not** in scope — it writes a keymap to a temp file rather than sizing an shm buffer, and folding it in would widen the helper to serve two jobs.

No behavior changes for a user. The win is that the portability guard and the `ftruncate` `EINTR` retry are written once and every backend gets them.

## Acceptance criteria

- [ ] One shared helper creates the shm fd; the three private copies are gone
- [ ] The shared helper keeps the `#ifdef __NR_memfd_create` guard and the `mkstemp` fallback, so screencopy gains the portability the other two already had
- [ ] `neru_wlr_create_keymap_fd` is left alone
- [ ] The new files sit in the right build-tagged slots per `internal/adapter/platform/linux/AGENTS.md`
- [ ] `just ci` and `just lint-cross` pass; the new `.c` compiles clean under `-Wall -Wextra -Wconversion`

## Blocked by

Nothing. Worth landing before #1478 adds a fourth call site.

y3owk1n:
## Description

This PR reworks how the Linux Wayland backends allocate the shared-memory
buffer they hand to the compositor. Three of them — the overlay, the wlroots
cursor-discovery surface, and the screencopy screen grab — each carried their
own hand-copied version of the same routine, and the copies had already
drifted apart. The two older ones fell back to a temporary file on kernels
whose headers predate `memfd_create`; the newest one, added with the wlroots
screencopy path, did not, so on such a target a screen capture failed where an
overlay draw succeeded. Nobody chose that difference — it is what copying a
routine by hand costs on the third repetition.

All three now share one routine. Screencopy gains the portability fallback the
other two already had, and the retry that keeps a signal arriving mid-allocation
from silently producing a short buffer is written once instead of three times.
Nothing else changes for a user: the buffers are the same size, the failure
paths return the same values, and the debug labels each backend puts on its
buffer are preserved.

Two deliberate consequences of the consolidation:

- The fallback path's temporary-file name is now the same for every caller
  rather than one per backend. The file is unlinked the instant it is created,
  so nothing can ever see the name.
- The fallback now refuses when it cannot remove that temporary file, where all
  three originals carried on regardless. That was the one case where screen
  content could land in a file that persists on disk under a name — small, but
  the whole point of the fallback is that it does not. Callers already handle
  the failure, and reaching it needs a `/tmp` that accepts a new file but will
  not let it be removed.

## Related Issues

Closes #1479

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix

## Cross-Platform Checklist

- [x] OS-specific files use correct build tags (e.g., `//go:build darwin`) — the
      new `.c`/`.h` pair sits in the single-platform `platform/linux` directory
      and compiles through the one `linux && cgo` unit, like every other `.c`
      there
- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) —
      N/A, this is internal C behind existing Go entry points; no port surface
      changed, so the existing `_nocgo.go` twins still cover it

## General Checklist

- [x] Tests added/updated for new or changed functionality — N/A, no behaviour
      changed and the code is C reached only from a live Wayland session; the
      compile-time gates below are what cover it
- [x] Documentation updated (if applicable) — N/A, no capability status or
      user-facing surface moved, so the ownership table in
      `docs/CROSS_PLATFORM.md` is unaffected

## Additional Context

Verification beyond the gate, since this is C that a macOS host never compiles:

- `just lint-cross` — clean (Linux amd64 with CGO on, which is what actually
  compiles these files).
- The new file compiles warning-free under `-Wall -Wextra -Wconversion
  -Wsign-conversion -Wpedantic` on GCC 14, exercised twice: once normally, and
  once with `__NR_memfd_create` forcibly undefined so the fallback branch gets
  the same sweep.

Two decisions worth a sentence each:

- The shared routine takes a `size_t`, which is the type all three call sites
  already computed. That puts the one signed conversion inside the routine
  instead of at two of the three calls, and makes a negative size
  unrepresentable at the boundary.
- The keymap-file helper next door in the wlroots client is left alone. It
  writes a keymap into a temporary file rather than sizing a buffer, and
  folding it in would widen the shared routine to serve two jobs.

Neither path sets close-on-exec — that is unchanged from before, and is now in
one place rather than three if it is ever worth revisiting.

### DIFF
--- a/internal/adapter/platform/linux/overlay_wayland.c
+++ b/internal/adapter/platform/linux/overlay_wayland.c
@@ -1,5 +1,6 @@
 #include "overlay_wayland.h"
 
+#include "shm_file.h"
 #include "wlr_protocol/fractional-scale-v1.h"
 #include "wlr_protocol/layer-shell.h"
 #include "wlr_protocol/viewporter.h"
@@ -7,8 +8,6 @@
 #include "wlr_protocol/xdg-shell.h"
 
 #include <cairo/cairo.h>
-#include <errno.h>
-#include <fcntl.h>
 #include <math.h>
 #include <poll.h>
 #include <stdio.h>
@@ -60,38 +59,6 @@ static const char *neru_modifier_name_from_keysym(xkb_keysym_t keysym) {
 	default:
 		return NULL;
 	}
-}
-
-// Create anonymous shared memory
-static int create_shm_file(off_t size) {
-	int ret, fd;
-
-#ifdef __NR_memfd_create
-	fd = syscall(__NR_memfd_create, "neru-overlay-shm", 0);
-	if (fd >= 0) {
-		do {
-			ret = ftruncate(fd, size);
-		} while (ret < 0 && errno == EINTR);
-		if (ret >= 0)
-			return fd;
-		close(fd);
-	}
-#endif
-
-	// Fallback if no memfd
-	char name[] = "/tmp/neru-shm-XXXXXX";
-	fd = mkstemp(name);
-	if (fd < 0)
-		return -1;
-	unlink(name);
-	do {
-		ret = ftruncate(fd, size);
-	} while (ret < 0 && errno == EINTR);
-	if (ret < 0) {
-		close(fd);
-		return -1;
-	}
-	return fd;
 }
 
 static void neru_layer_surface_configure(
@@ -600,7 +567,7 @@ static int neru_create_single_buffer(
     NeruWaylandOverlay *overlay, NeruWaylandOverlayScreen *scr, int buf_idx, int buf_width, int buf_height, int stride,
     double scale) {
 	size_t buf_size = (size_t)stride * (size_t)buf_height;
-	int fd = create_shm_file(buf_size);
+	int fd = neru_shm_file_create("neru-overlay-shm", buf_size);
 	if (fd < 0)
 		return -1;
 
--- a/internal/adapter/platform/linux/shm_file.c
+++ b/internal/adapter/platform/linux/shm_file.c
@@ -1,0 +1,60 @@
+#include "shm_file.h"
+
+#include <errno.h>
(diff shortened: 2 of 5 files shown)


y3owk1n:
Good catch on the unlink — fixed in 02c2a49d. The three originals all ignored what `unlink` returned, but the shared helper's comment now states the anonymity as a guarantee, so it should hold it: a failed unlink closes the descriptor and returns -1, which every caller already degrades on.

Issue #1492, opened 2026-08-12
perf(linux): the grid cell-selection keystroke repaints the shared surface twice


y3owk1n:
## What to build

The keystroke that chooses a cell in grid mode repaints the whole shared Cairo surface twice on Linux. `internal/app/modes/grid.go` calls `showGridSubgrid(cell)` and then `refreshGridVirtualPointer()` back to back (lines 220-221 and 233-234); the pointer's position has just changed, so the second call repaints the subgrid the first one only just painted. `internal/app/modes/cleanup.go` has the same pair — `updateGridMatches("")` then hiding the pointer, immediately before `clearOverlayFrame`. Arrow keys inside a subgrid go through the same callback, so a held key pays it per repeat.

This is the deliberate limitation PR #1488 disclosed rather than a regression it hid: on Linux the pointer shares one surface with the grid, so moving it repaints the grid. The equality guard in `SetGridPointer` already keeps the common narrowing keystroke at the one repaint it always had, so this is scoped to the keys that move the selection. macOS pays a cheap pointer-only layer update instead, so it is a genuine per-platform asymmetry on a per-keystroke path — and *latency is the product*.

The fix is the subgrid call carrying the pointer the way the recursive-grid frame already does, so one call paints both. That is a `ports.OverlayPort` / `manager.Interface` change on a per-keystroke path, which is exactly why #1488 did not take it on inside its own scope: ADR 0003 keeps grid mode's per-keystroke updates as plain calls rather than frames, so widening the call is the move, not converting grid mode to frames.

**Measure before and after.** #1211 set the precedent of quoting real numbers for a per-keystroke path (grid narrowing unchanged; recursive grid +1 allocation of 128 B, no measurable time cost). A change here that does not report its numbers has not shown it helped.

## Acceptance criteria

- [ ] Choosing a cell repaints the shared surface once, not twice
- [ ] The teardown path in `cleanup.go` does not repaint before clearing either
- [ ] Held arrow-key repeat inside a subgrid pays one repaint per press
- [ ] The narrowing keystroke is still deduped and no slower than today
- [ ] Grid mode's per-keystroke updates are still plain calls, not frames (ADR 0003)
- [ ] Before/after numbers for the selection keystroke are in the PR
- [ ] `just ci` passes, plus `just lint-cross`

## Blocked by

Nothing.

Issue #1491, opened 2026-08-12
fix(linux): the main grid cells reappear behind an open subgrid on the next keystroke


y3owk1n:
## What to build

On Linux, opening a subgrid in grid mode looks right until the next keystroke, at which point the parent grid's cells reappear behind the subgrid.

Two functions in `internal/adapter/overlay/linux/overlay_shared_cgo.go` disagree about what an open subgrid should show:

- `showSubgrid` (the open) paints the subgrid **alone** — `paintSubgridContent(cell)` and nothing else.
- `redrawGrid` (every narrowing keystroke after it) paints **all** the parent cells, then draws the subgrid over them when `currentSubgrid != nil`.

So the first keystroke after an open changes what is on screen without the user asking it to. Pre-existing, Linux only, found while shipping #1463 and left alone because changing it alters an existing path's visible behaviour.

**Settle which one is right against macOS, not by preference.** macOS is the reference implementation, so the ticket's first job is to look at what the darwin grid overlay does with the parent cells while a subgrid is open, and make both Linux functions agree with that answer. Do not pick whichever is cheaper to implement — a wrong choice here is a parity divergence that reads as correct.

## Acceptance criteria

- [ ] What the darwin overlay shows behind an open subgrid is established and stated in the PR
- [ ] `showSubgrid` and `redrawGrid` agree with each other and with macOS
- [ ] No keystroke changes what is on screen except through what the user typed
- [ ] A test pins it, so the two paths cannot drift apart again
- [ ] Verified on X11 and a wlroots compositor
- [ ] `just ci` passes, plus `just lint-cross` (this file is `linux && cgo`, which `just lint` cannot see)

## Blocked by

Nothing. **But** it touches the same two functions as the sibling ticket on `showSubgrid` clearing before `beginFrame` — run them sequentially, not in parallel.

y3owk1n:
PR: https://github.com/y3owk1n/neru/pull/1509 (branch `fix/linux-subgrid-parent-cells`)

Issue #1463, opened 2026-08-11
feat(linux): draw the virtual-pointer indicator in grid mode, as recursive grid already does


y3owk1n:
## What to build

The virtual-pointer indicator is drawn on Linux in **recursive grid** and is a no-op in **grid**. macOS draws it in both. So the same indicator, in the same overlay system, on the same platform, appears in one mode and not its sibling — which reads to a user as a bug in grid mode rather than a missing platform feature.

`CONTEXT.md` defines an Indicator as "a small overlay that tracks the cursor and reports state, independent of the active mode's own drawing". Independent of the mode is the point: an indicator that appears in one mode and not another is not behaving like an indicator.

The drawing already exists for recursive grid on both Linux backends. This is wiring it into grid mode's frame, not writing a renderer.

Check the `UseVirtualPointerOverlay` path on the darwin manager while you are here — it has no Linux counterpart, and whether that matters depends on how grid mode requests the indicator.

## Acceptance criteria

- [ ] Grid mode draws the virtual-pointer indicator on X11, wlroots and KDE
- [ ] It behaves as it does in recursive grid — same style resolution, same tracking, same teardown
- [ ] It is torn down cleanly when the mode exits, including mid-animation
- [ ] The **Grid / Virtual pointer indicator** row in the mode-coverage table is updated
- [ ] `docs/CROSS_PLATFORM.md` Known Gaps entry 10 is removed
- [ ] `just ci` passes

## Blocked by

Nothing.

## Parent

#1466 — Linux parity

PR #1507, opened 2026-08-12
fix(linux): show the subgrid instead of a stale frame when a cell is picked

### LINKED ISSUE #1490: fix(linux): showSubgrid clears before it begins the frame, and cancels an animation with renderMu held
## What to build

Two defects in one ten-line function — `sharedOverlay.showSubgrid` in `internal/adapter/overlay/linux/overlay_shared_cgo.go`. Both were found while shipping #1463, which deliberately routed its new code around them rather than change an existing path inside that ticket's scope.

**1. It clears before it begins the frame.** The sequence today is `ensureBuffers()` → `clear()` → `beginFrame()`. On Wayland, `beginFrame` is what selects the writable buffer, so clearing before it wipes the buffer that is currently on screen and leaves the one about to be shown holding whatever it last held. This is the exact shape Greptile caught as a P1 in PR #1488 — in code that had just been written — and it is already fixed there: both `redrawGrid` and the new `repaintGridSurface` do `beginFrame()` then `surfaceClear()`. `showSubgrid` is the last caller with the order the wrong way round, and unlike the latent one below it runs on **every subgrid open on Wayland**.

**2. It reaches `cancelAnimation` with `renderMu` held.** `Manager.ShowSubgrid` takes `renderMu`, then `showSubgrid` calls `clear()`, which begins with `cancelAnimation()`. That function's own doc comment states the contract it is being called against: *"The caller must not hold renderMu: the goroutine takes it on every frame and would never reach the stop signal."* It only bites while a shared-surface animation is painting at the moment a subgrid opens — which the grid frame coming up has usually already cancelled — so this is latent rather than live. It is still a documented contract violated in-tree.

The fixes interact, which is why they are one ticket rather than two: `clear()` does four things (cancel the animation, reset `hasLast`/`lastHints`, clear the surface) and the fix splits them across the `beginFrame` boundary. `HideHintSearchInput` already shows the shape for hoisting the cancel above the lock.

## Acceptance criteria

- [ ] `showSubgrid` begins the frame before clearing it, as `redrawGrid` and `repaintGridSurface` do
- [ ] The animation cancel happens with `renderMu` released — `HideHintSearchInput` is the precedent
- [ ] The state `clear()` resets (`hasLast`, `lastHints`) is still reset, and at a point where resetting it is correct
- [ ] A test pins the frame-primitive order, as `TestLinuxOverlayManager_DrawGridPointer_BeginsTheFrameBeforeClearingIt` does for the #1463 path — `recordingSurface` already records that order
- [ ] Verified on a wlroots compositor: opening a subgrid shows the subgrid, not a stale frame
- [ ] `just ci` passes, plus `just lint-cross` (this file is `linux && cgo`, which `just lint` cannot see)

## Blocked by

Nothing. **But** it touches the same two functions as the sibling ticket on `redrawGrid` painting the main cells under an open subgrid — run them sequentially, not in parallel.

y3owk1n:
## Description

This PR fixes opening a subgrid on Wayland showing a stale frame. Picking a cell in grid mode cleared the overlay before the compositor had handed over a buffer to draw into, so the clear landed on the frame already on screen and the frame that came up next still held the previous one — a leftover grid where the subgrid should be, on every subgrid open on a wlroots compositor. It now clears the frame it is about to show, the way the grid redraw and the pointer repaint already do.

It also fixes a hang that was waiting to happen: opening a subgrid while the overlay was mid-animation stopped that animation from inside the render lock, and the animation needs that same lock to finish a frame — so the two would have waited on each other with the grid still on screen. Stopping it happens before the lock now, as hiding the hint search input already does.

Live verification on a wlroots compositor is outstanding — this was written on macOS, which has no compositor to open a subgrid on, so the in-repo evidence is a test that pins the order the frame primitives run in (the same shape as the one added for the grid pointer) plus a second one pinning that the animation is stopped with the render lock released. A maintainer on wlroots should confirm the screen.

## Related Issues

Closes #1490. Found while shipping #1463, which routed around both defects rather than change an existing path inside that ticket's scope.

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no new platform surface: this changes the order of calls inside an existing Linux backend method.

## General Checklist

- [x] Documentation updated (if applicable) — no capability status or user-facing surface changed; the overlay guide's lock-topology rule now states the cancel-before-the-lock contract this restores, instead of it living only in the comment on each call site.

## Additional Context

`just lint-cross` was run as well as `just ci`, both green — this file is `linux && cgo`, which the host lint cannot see. The Linux test suite was also run in the CI-equivalent container, with `-race` and with CGO off.

Two things worth a reviewer's eye:

- The two records the old clear dropped (the recursive-grid animation's last bounds, and the hint labels the search badge repaints over) are dropped **after** the clear rather than before it, which is the opposite order from the two neighbouring draws that do the same thing. It is deliberate: if the frame cannot be begun nothing was painted, so the old frame is still on screen and both records still describe it truthfully. Dropping them above the guard would claim a screen that was never drawn.
- Stopping the animation before taking the lock opens the same short window every other canceling call here already has — the backend pointer is read once for the cancel and re-read under the lock for the dispatch. The subgrid open is now in the shutdown race table alongside those calls, so that window is covered.

Neither sibling ticket is absorbed: the grid redraw's cell painting (#1491) and the double repaint on the selection keystroke (#1492) are untouched.


y3owk1n:
The one red check — `test (windows-latest)` — never reached the test run. Chocolatey's package feed is intermittently returning 503/504 for `just` today, so the setup step installs nothing and `just test-ci` exits 127 with `just: command not found`. It has been re-run three times with the same result, while `build`/`lint`/`vet`/`vuln` on the same runner and the same commit went green once the feed recovered for them.

Nothing in this branch touches Windows: the Windows test binaries compile (`just ci` includes the Windows test-binary compile, green locally), and the changed code is behind `//go:build linux && cgo`. Worth another re-run when the feed settles.

PR #1523, opened 2026-08-14
fix(logging): keep the daemon log per-user and out of its own content

y3owk1n:
## Description

This PR moves the daemon's log output out of `/tmp` and into the user's own log
directory, and stops three log lines from writing down content instead of facts
about it.

On macOS the login agent used to send the daemon's standard output and standard
error to `/tmp/neru.log` and `/tmp/neru.err.log`. `/tmp` is shared between every
account on the machine, and the standard-output copy was a duplicate of the
rotated log file Neru already writes and documents. Standard output is now left
unredirected, and standard error goes to `~/Library/Logs/neru/daemon.err.log`,
beside `app.log`, so a crash or a startup failure raised before the log file is
open still lands somewhere readable. The plist the CLI generates, the template
shipped for a hand install, the macOS installer script and both Nix modules all
move together, and `neru services install` now creates the log directory before
launchd is asked to write into it.

Separately, three log fields carried the thing itself rather than a fact about
it: a failed `exec` sequence step logged the whole command line and its combined
output at error level, and the two config-set paths logged the value beside the
key at info level. Configuration values can be shell command lines, and the
output is whatever the user's own command printed — neither belongs in a log
file, per the logging rule in `AGENTS.md`. Those sites now log a length, a byte
count and an exit code; the error handed back to whoever ran the command is
unchanged, so no diagnostic detail is lost where it was actually being read.

## Related Issues

None.

## Target Platform

- [x] macOS
- [ ] Linux
- [ ] Windows

## Type of Change

- [x] `fix` — Bug fix

## Cross-Platform Checklist

- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no port or capability changes; the launchd work is inside the existing `//go:build darwin` service file and the logger's path helper already covers all three platforms.

## General Checklist

## User-facing surface

No config option, command or flag changes. What changes is where the macOS login
agent's output goes:

| | Before | After |
| --- | --- | --- |
| stdout | `/tmp/neru.log` | not redirected — `~/Library/Logs/neru/app.log` already holds every log line |
| stderr | `/tmp/neru.err.log` | `~/Library/Logs/neru/daemon.err.log` |

`neru services install` creates `~/Library/Logs/neru` itself, so a fresh install
captures its first launch. The nix-darwin module redirects neither stream: that
agent is installed system-wide and runs as whichever user logs in, so no per-user
path can be written into it at build time, and the shared one was the thing being
removed. The home-manager module names the user's own path.

## Potentially breaking

An agent installed by an earlier version keeps its old plist — installing over an
existing one is refused rather than silently rewritten, which is deliberate. A
user who wants the new location runs `neru services uninstall && neru services
install` and deletes the two files left in `/tmp`. `docs/CLI.md` now says so under
`neru services`. Anyone tailing `/tmp/neru.log` out of habit should tail
`~/Library/Logs/neru/app.log` instead, which is the path the troubleshooting docs
have always given.

## Additional Context

Two guardrails keep both halves from drifting back, since both regressions are
the kind that read like helpful diagnostics in review:

- no log call above debug level may name a field `value`, `output`, `cmd` or
  `command` — the correct substitutes (`cmd_length`, `output_bytes`) spell
  themselves, since a length is not the content;
- no file that decides where the daemon's output goes, in any language, may name
  a shared temp directory. A file that redirects nothing drops out of the subject
  set; adding a redirect back puts it under the rule again.

The `exec` step's debug-level log of the command string is untouched: debug is
opt-in, and this change is about what a default install writes down.

One thing found while writing the second path into the plist and fixed here too:
the paths were being substituted into the XML unescaped, so a binary living under
a directory called `A&B` — legal on macOS — produced a plist launchctl refuses to
load, after the file had already been written, which then made the next install
refuse as well. Both substitutions are escaped now.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -1533,6 +1533,17 @@ systemd user unit on Linux; where the Linux unit is written, what it contains,
 and what happens on a machine booted by another init system are in
 [LINUX_SETUP.md](LINUX_SETUP.md#systemd-user-service).
 
+The macOS agent leaves the daemon's standard output alone — the rotated log file
+already holds every log line — and sends its standard error to
+`~/Library/Logs/neru/daemon.err.log`, beside that log file, where a crash or a
+failure raised before the log file is open ends up. Both stay inside the user's
+own log directory; neither goes to a shared path.
+
+An agent installed by an earlier version wrote those two streams to `/tmp`
+instead, and installing over it is refused rather than silently rewritten. To
+move an existing one, run `neru services uninstall && neru services install`,
+then delete the files it left behind at `/tmp/neru.log` and `/tmp/neru.err.log`.
+
 `status` reports a machine where the service was never installed as exactly
 that, rather than failing.
 
--- a/internal/adapter/logger/logger_path.go
+++ b/internal/adapter/logger/logger_path.go
@@ -9,6 +9,20 @@ import (
 // defaultLogFilePath returns the platform default neru log file path when
 // [logging].log_file is empty.
 func defaultLogFilePath() (string, error) {
+	logDir, err := DefaultLogDir()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(logDir, "app.log"), nil
+}
+
+// DefaultLogDir returns the per-user directory neru writes its logs to. It is
+// exported because the daemon's log file is not the only thing that belongs
+// there: a service definition redirecting the daemon's stderr needs the same
+// directory, and a second spelling of it would be a second answer to where
+// neru's logs live.
+func DefaultLogDir() (string, error) {
 	homeDir, err := os.UserHomeDir()
 	if err != nil {
 		return "", err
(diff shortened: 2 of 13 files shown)


y3owk1n:
`vuln (ubuntu-latest)` is red here for a reason unrelated to this branch, and
`main` would go red the same way on its next run.

It reports `GO-2026-5972` (a recursion-depth fix in the standard library's
`encoding/asn1`, fixed in Go 1.26.6) reached through `sync.Once.Do` from
`internal/app/lifecycle.go` — untouched by this PR, which changes no dependency,
no `go.mod` and no toolchain pin. Linux is the only leg that fails because it is
the only one whose `crypto/x509` root loading goes through `asn1`; macOS and
Windows use their platform verifiers, so the call graph does not exist there.

Linux and Windows install Go from the `go 1.26.4` line in `go.mod`, so clearing
it means bumping that (and devbox, which supplies the macOS toolchain) — a
separate change from this one.

PR #1656, opened 2026-09-09
fix(linux): make a hotkey click on Wayland land where the cursor is, unmodified

y3owk1n:
## Description

This PR fixes `action left_click` fired from a global hotkey on Wayland. Two things went wrong on Hyprland with `"Alt+Space" = "action left_click"` and no mode active.

The click landed on a stale point. The Wayland adapter caches the cursor position, and only mode activation refreshed it, so moving the mouse by hand and then firing the hotkey clicked wherever the cache last pointed. Actions now resync the cache before resolving their target, under the same 150ms budget modes use. macOS and X11 query the OS directly and are unchanged.

The click arrived as alt+click. The evdev proxy withholds the chord's key but forwards its modifier, and a Wayland pointer event carries whatever the seat's keyboards hold. The proxy now releases the physically held modifiers on its uinput keyboard around a button event and presses them again afterwards, the same rule X11 and Windows already apply from the live key state (footnote 7 in `docs/CROSS_PLATFORM.md`). A physical release during the click is swallowed, and the re-press is followed by the symbol-less tap that keeps a modifier-alone launcher shortcut from firing. Without a forwarding proxy (no `/dev/uinput`, or the wl-keyboard fallback) nothing is lifted and the click stays modified, which the docs now state.

How to test on Hyprland: bind `"Alt+Space" = "action left_click"` under `[hotkeys]`, restart the daemon, move the mouse by hand, press Alt+Space. The click should land under the pointer with no Alt. Keep Alt held afterwards and press Tab: Alt+Tab should still work.

## Related Issues

None.

## Target Platform

- [ ] macOS
- [x] Linux
- [ ] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

The nocgo stubs for `LiftHeldModifiers` / `RestoreLiftedModifiers` report nothing lifted rather than `CodeNotSupported`: the caller degrades to a modified click either way, and an error would abort the click.

## General Checklist

- [x] `just ci` passes
- [x] PR title is a conventional commit subject written for users

## Additional Context

The cursor resync costs one `hyprctl cursorpos` round trip per idle action on Hyprland, and a layer-shell discovery pass on Sway and niri. In-mode actions with a selection point never reach it. The modifier lift only runs when a modifier is physically held, so plain clicks pay nothing.

### DIFF
--- a/docs/CROSS_PLATFORM.md
+++ b/docs/CROSS_PLATFORM.md
@@ -171,7 +171,7 @@ answer.
 | **Keymap learns the focused app** | ✅ published by the watcher | ✅ published by the watcher | ✅ published by the watcher | ✅ published by the watcher | ✅ published by the watcher |
 | **Cursor position**           | ✅ `CGEventGetLocation`  | ✅ `XQueryPointer`     | ✅ `hyprctl` on Hyprland, else sync-surface cache | ✅ sync-surface cache | ✅ `GetCursorPos` |
 | **Cursor move**               | ✅ `CGEventPost` ([`postMouseMoveLocked`](../internal/adapter/platform/darwin/accessibility_mouse_darwin.m)) | ✅ XTest (`XTestFakeMotionEvent`) | ✅ `zwlr_virtual_pointer` | ✅ libei                | ✅ `SetCursorPos`            |
-| **Mouse buttons / drag**      | ✅ `CGEventPost`         | ✅ XTest ⁷             | ✅ `zwlr_virtual_pointer`    | ✅ libei                | ✅ `SendInput` ⁷             |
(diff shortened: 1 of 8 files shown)


y3owk1n:
Addressed: the nocgo stubs now return `derrors.CodeNotSupported`. The caller is cgo-only and already degrades on any error to a click that keeps the held modifiers.

y3owk1n:
Addressed both comments outside the diff: a failed press now restores the modifiers it lifted, and a release restores only when it ends the last held button, so an unrelated `mouse_up` no longer re-modifies a drag in progress.

y3owk1n:
Addressed: a failed release of the last held button now restores the lifted modifiers as well, keeping the button recorded so the idle cleanup can retry the release.

y3owk1n:
Addressed: a failed release now re-records the held button without its modifiers, since the deferred cleanup already released them, so the idle retry does not release them a second time.

y3owk1n:
Addressed: a failed release now records only the modifiers whose own release failed, so the idle retry lets go of exactly those and nothing else.

PR #1657, opened 2026-09-09
fix(overlay): draw only hint labels and badges bold on Windows and Linux

y3owk1n:
## Description

This PR fixes the font weight of overlay text on Windows and Linux. Every string the overlay drew on those platforms was bold, regardless of surface, so grid labels, sub-key previews, the hint search input and the virtual pointer glyph came out heavier than the same surfaces on macOS.

Text now uses the weight the macOS adapter uses per surface. Hint labels, the mode and sticky-modifier badges and the monitor-select label stay bold. Grid, subgrid, recursive-grid labels, sub-key previews, the search input, the monitor-select subtitle and the virtual pointer glyph draw at regular weight. Both Windows renderers and both Linux backends follow the same table. No config option changes.

## Related Issues

None. Reported by a Windows user on the nightly build.

## Target Platform

- [ ] macOS
- [x] Linux
- [x] Windows

## Type of Change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Cross-Platform Checklist

- [x] Stub implementations added for other platforms (returning `CodeNotSupported`) — N/A, no port contract changed; only package-private primitives gained a parameter

## General Checklist

- [x] `just ci` passes — N/A, ran the same recipes individually on this host: `fmt-check`, `lint`, `vet`, `test`, `check-cross`, plus `lint-cross` for the Linux cgo build in Docker
- [x] Documentation updated (if applicable) — N/A, font weight is not a documented capability
- [x] PR title is a [conventional commit](https://www.conventionalcommits.org/) subject written for users

## Screenshots / Recordings

Not captured. This host is macOS, where nothing changes. A before/after from a Windows or Linux nightly would be welcome.

## Additional Context

The per-surface weight table was read off the macOS adapter, which resolves fonts with an explicit bold flag per call site. The `test-linux` Docker run has one pre-existing failure unrelated to this change, a GNOME focused-app test that needs `dbus-launch` in the image.

### DIFF
--- a/internal/adapter/overlay/linux/corner_radius_test.go
+++ b/internal/adapter/overlay/linux/corner_radius_test.go
@@ -57,6 +57,7 @@ type recordedText struct {
 	center     image.Point
 	fontSize   float64
 	color      uint32
+	bold       bool
 }
 
 func (s *recordingSurface) alive() bool { return true }
@@ -114,7 +115,7 @@ func (s *recordingSurface) hintBadgePrim(
 
 func (s *recordingSurface) textPrim(
 	text, fontFamily string, centerX, centerY, fontSize float64, color uint32,
-	_ bool,
+	bold bool,
 ) {
 	s.texts = append(s.texts, recordedText{
 		text:       text,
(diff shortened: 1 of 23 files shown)


y3owk1n:
Addressed the font-weight coverage comment: the recording fakes on both platforms now keep the bold flag, and a test per platform pins hint labels bold and the search input and grid labels regular.

